PR #59, opened 2020-10-27
CBAPI-1508: Add Asynchronous Query Support and Demonstration on DeviceSearchQuery

abowersox-cb:
## Pull request checklist

https://jira.carbonblack.local/browse/CBAPI-1508

## Pull Request Description

Asynchronous query support added to CBC SDK:
* Support for asynchronous executor object in CBCloudAPI
* New AsyncQueryMixin to be added to query objects for asynchronous query support
* Mixin added to DeviceSearchQuery and support added
* list_devices.py example updated to optionally use async query (execute_async() method) via --asynchronous or -Y command line flag

- [ ] Yes
- [x] No

## How Has This Been Tested?

Unit tests have been added for specific functionality. In addition, list_devices.py has been run both with and without the asynchronous option and the outputs have been verified to be equivalent.

### DIFF
--- a/src/cbc_sdk/rest_api.py
+++ b/src/cbc_sdk/rest_api.py
@@ -11,6 +11,8 @@
 # * WARRANTIES OR CONDITIONS OF MERCHANTABILITY, SATISFACTORY QUALITY,
 # * NON-INFRINGEMENT AND FITNESS FOR A PARTICULAR PURPOSE.
 
+"""Definition of the CBCloudAPI object, the core object for interacting with the Carbon Black Cloud SDK."""
+
 from cbc_sdk.connection import BaseAPI
 from cbc_sdk.errors import ApiError, CredentialError, ServerError
 from cbc_sdk.live_response_api import LiveResponseSessionManager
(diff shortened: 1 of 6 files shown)


src/cbc_sdk/rest_api.py
```diff
@@ -250,7 +273,7 @@ def bulk_threat_dismiss(self, threat_ids, remediation=None, comment=None):
         """
         return self._bulk_threat_update_status(threat_ids, "DISMISSED", remediation, comment)
 
-    # ---- ThreathHunter
+    # ---- ThreatHunter
```
avanbrunt-cb, line 276: This should be Enterprise EDR

abowersox-cb: Heh, I fixed the typo, which is the only reason this showed up. :) Perhaps I should add another ticket to do a sweep of the comments for any outdated terminology?

avanbrunt-cb (approved):
LGTM. Are there any other failure tests cases around async that might happen? or are those hard to trigger?

abowersox-cb:
> Are there any other failure tests cases around async that might happen? or are those hard to trigger?

Well, the implementation as it exists is pretty simple and not particularly error-prone in terms of the code that was written; that said, _anything_ involving multiple threads is going to be hard to trigger failures in.

llyon-cb (approved):

PR #60, opened 2020-10-30
Enriched events search support

mrangelov-cb:
## Pull request checklist

- [ ] Yes
- [x] No


src/cbc_sdk/enterprise_edr/base.py
```diff
@@ -197,20 +197,19 @@ def process_pids(self):
             return None
 
 
-class Event(UnrefreshableModel):
-    """Events can be queried for via `CBCloudAPI.select` or an already selected process with `Process.events`."""
-    urlobject = '/api/investigate/v2/orgs/{}/events/{}/_search'
-    validation_url = '/api/investigate/v1/orgs/{}/events/search_validation'
-    default_sort = 'last_update desc'
-    primary_key = "process_guid"
+class EnrichedEvent(UnrefreshableModel):
```
avanbrunt-cb, line 200: This is an endpoint_standard model not enterprise_edr

src/cbc_sdk/enterprise_edr/base.py
```diff
@@ -612,3 +627,179 @@ def _perform_query(self):
             results["incomplete_results"] = result["incomplete_results"]
 
         return results
+
+
+class AsyncEventQuery(Query):
```
avanbrunt-cb, line 632: This should probably be called AsyncEnrichedEventQuery

src/tests/unit/enterprise_edr/test_enterprise_edr_enriched_events.py
```diff
@@ -0,0 +1,64 @@
...
+    for event in events:
+        assert event.device_name is not None
+        assert event.enriched is not None
+
+def test_enriched_event_query_implementation(cbcsdk_mock):
+    """Testing EnrichedEvent querying with where()."""
+    api = cbcsdk_mock.api
+    event_id = '27a278d5150911eb86f1011a55e73b72'
+    event = api.select(EnrichedEvent).where(f"event_id:{event_id}")
+    assert isinstance(event, AsyncEventQuery)
```
avanbrunt-cb, line 64: You should also verify that the where property is set correctly inside the AsyncEventQuery as well

src/cbc_sdk/enterprise_edr/base.py
mrangelov-cb: Changed to endpoint_standard.

src/cbc_sdk/enterprise_edr/base.py
mrangelov-cb: Done.

src/tests/unit/enterprise_edr/test_enterprise_edr_enriched_events.py
mrangelov-cb: That one corrected as well.

src/cbc_sdk/enterprise_edr/base.py
mrangelov-cb: Changed as requested

src/cbc_sdk/enterprise_edr/base.py
mrangelov-cb: Done.

src/tests/unit/enterprise_edr/test_enterprise_edr_enriched_events.py
mrangelov-cb: Done.

llyon-cb (approved):
Looks good to me!

abowersox-cb (approved):

avanbrunt-cb (approved):
LGTM the branch just needs updated

src/cbc_sdk/enterprise_edr/base.py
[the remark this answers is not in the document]
mrangelov-cb: Changed to EnrichedEventQuery

src/cbc_sdk/endpoint_standard/base.py
```diff
@@ -368,3 +384,180 @@ def _search(self, start=0, rows=0):
...
+        result_url_template = "/api/investigate/v2/orgs/{}/enriched_events/search_jobs/{}/results".format(
+            self._cb.credentials.org_key,
+            self._query_token
+        )
+        query_parameters = {}
+        while still_fetching:
+            result_url = '{}?start={}&rows={}'.format(
+                result_url_template,
+                current,
+                10  # Batch gets to reduce API calls
```
cnotin, line 546: Why not more? Larger batches are faster :)

cnotin:
Thank you very much for this implementation that will help me a lot!
I tested it and it works fine :)
I just added one suggestion regarding the batch size: https://github.com/carbonblack/carbon-black-cloud-sdk-python/pull/60/files#r519823926

cnotin:
Also, it would be really nice to have support for the "time_range" parameter which is available in this API according to its doc. I think its implementation can be copied from the one in the enterprise_edr module:
https://github.com/carbonblack/carbon-black-cloud-sdk-python/blob/df6af9a0c59dccf8c1676ae390848efbb76136d3/src/cbc_sdk/enterprise_edr/base.py#L393-L410

Let me know if you want me to suggest a PR!

src/cbc_sdk/endpoint_standard/base.py
cnotin: Actually, there is `self._batch_size = 100` so it could be used here instead of the hardcoded value

mrangelov-cb: Done, changed as requested.

src/cbc_sdk/endpoint_standard/base.py
```diff
@@ -417,6 +418,46 @@ def _get_query_parameters(self):
...
+            start (str in ISO 8601 timestamp): When to start the result search.
+            end (str in ISO 8601 timestamp): When to end the result search.
+            window (str): Time window to execute the result search, ending on the current time.
+                Should be in the form "-2w", where y=year, w=week, d=day, h=hour, m=minute, s=second.
+        Note:
+            - `window` will take precendent over `start` and `end` if provided.
+        Examples:
+            query = api.select(Event).set_time_range(start="2020-10-20T20:34:07Z")
+            second_query = api.select(Event).set_time_range(start="2020-10-20T20:34:07Z", end="2020-10-30T20:34:07Z")
+            third_query = api.select(Event).set_time_range(window='-3d')
```
cnotin, line 446:
```suggestion
            query = api.select(EnrichedEvent).set_time_range(start="2020-10-20T20:34:07Z")
            second_query = api.select(EnrichedEvent).set_time_range(start="2020-10-20T20:34:07Z", end="2020-10-30T20:34:07Z")
            third_query = api.select(EnrichedEvent).set_time_range(window='-3d')
```
Adapt after copy/paste ;)

mrangelov-cb: Uh oh...fixed :)

src/cbc_sdk/endpoint_standard/base.py
```diff
@@ -368,3 +384,225 @@ def _search(self, start=0, rows=0):
...
+
+    def set_rows(self, rows):
+        """
+        Sets the 'rows' query body parameter, determining how many rows of results to request.
+        Args:
+            rows (int): How many rows to request.
+        """
+        if not isinstance(rows, int):
+            raise ApiError(f"Rows must be an integer. {rows} is a {type(rows)}.")
+        self._batch_size = rows
```
cnotin, line 429: I think we must distinguish between the total number of rows (parameter in the POST request to create the search job) to get, and the batch size (passed as a GET parameter in the /results requests to get results) for each retrieved page.
Here both are tied and if I use `set_rows(10000)` I may have an issue because I can request 10k total results, but will it be nice from the API to download 10k at once? There is no pagination anymore
It depends on how you prefer us to hit the API :)

cnotin:
Thank you @mrangelov-cb for your work! I appreciate a lot :)

Regarding set_time_range, I think one more thing is missing for it to work (sorry I didn't mention it earlier):
https://github.com/carbonblack/carbon-black-cloud-sdk-python/blob/df6af9a0c59dccf8c1676ae390848efbb76136d3/src/cbc_sdk/enterprise_edr/base.py#L431-L432

PS: I think some changes can be backported to the code in enterprise_edr, from which some of this implementation comes from

src/cbc_sdk/endpoint_standard/base.py
cnotin: You can close this yourself ;)

src/cbc_sdk/endpoint_standard/base.py
cnotin: Eheh ;) You can close this too!

src/cbc_sdk/endpoint_standard/base.py
cnotin: My suggestion would be to have a default batch size, of 50 or 100 for example, for page retrieval and perhaps keep it internal like in #69
Then expose a public `rows` parameter, as already done, which only allows to set the total number of results desired

mrangelov-cb: OK, switching to using the rows argument to the method for this. Also set the API defaults and we now check if we pass >10K API limit to set_rows.

src/cbc_sdk/endpoint_standard/base.py
```diff
@@ -415,6 +415,10 @@ def _get_query_parameters(self):
         """Need to override base class implementation as it sets custom (invalid) fields"""
         args = self._default_args.copy()
         args['query'] = self._query_builder._collapse()
+        if self._time_range:
+            args["time_range"] = self._time_range
+        if self._time_range:
+            args["time_range"] = self._time_range
```
cnotin, line 421: Oops duplicated lines ;)

src/cbc_sdk/endpoint_standard/base.py
cnotin: Good idea to check the max yes!

src/cbc_sdk/endpoint_standard/base.py
mrangelov-cb: Fixed :)

src/cbc_sdk/endpoint_standard/base.py
cnotin: Hum I'm still not very fan of the default of 10 but as you wish :)
I have a case where it sometimes uses 0 so it goes in an infinite loop... i'm trying to investigate

cnotin: Ok I found the issue, it comes from here:
https://github.com/carbonblack/carbon-black-cloud-sdk-python/blob/8e276bcc032115c66008e9faa3c7073806fb54e5/src/cbc_sdk/base.py#L724-L725
`numrows` is default so 0 which is passed to `EnrichedEventQuery._search`

mrangelov-cb: I am in a bit of confusion as to how to handle that. I'm thinking about introducing a check if it is zero and in that case, set it to API default. But this also doesn't look quite clean to me. Maybe introducing a second argument to set_rows to set the pagination batch size would be better. Still not sure. What do you think would be better and more intuitive?

mrangelov-cb: OK, I split that into set_rows() to set the total result rows and set_batch_size to set the batch size as per a paginated results call

avanbrunt-cb: Don't add `set_batch_size` as this already exists on the paginated query object which your class inherits https://github.com/carbonblack/carbon-black-cloud-sdk-python/blob/8e276bcc032115c66008e9faa3c7073806fb54e5/src/cbc_sdk/base.py#L728

 As for the infinite loop, there shouldn't be an infinite loop with the following lines in the code as this will break the loop. https://github.com/carbonblack/carbon-black-cloud-sdk-python/blob/CBPAPI-2025_enriched_events/src/cbc_sdk/endpoint_standard/base.py#L632-L633

avanbrunt-cb: You can also remove the following line https://github.com/carbonblack/carbon-black-cloud-sdk-python/blob/CBPAPI-2025_enriched_events/src/cbc_sdk/endpoint_standard/base.py#L398 as this is set by the super call to the parent inherited object here https://github.com/carbonblack/carbon-black-cloud-sdk-python/blob/CBPAPI-2025_enriched_events/src/cbc_sdk/endpoint_standard/base.py#L287

mrangelov-cb: Oops, didn't notice it has implementation in the base class...
Anyway, in the API documentation for that route they say the default value is 10 while in the Query class that we inherit it is initialized to 100. Still unsure what to do though as 10 is the default as per the REST API, but it doesn't necessarily need to be the CBC SDK default. Which would be better?

avanbrunt-cb: I would go with 100 since the 10 for the API is the default for the UI

mrangelov-cb: OK, removing set_batch_size and defaulting to 10 in the constructor.

avanbrunt-cb: No default to 100 we can be more optimal from a developer's perspective 10 is only needed for the UI

mrangelov-cb: I meant removing both the set_batch_size and defaulting to 10. Anyway, looks like I am too sleepy already, will review the changes and push tomorrow.

mrangelov-cb: OK, removed.

cnotin: FYI I just tested with the latest code as of now, and it works very nicely :)
I can use set_rows() to define how many results in total I want, and by default it fetches via pages of 100 rows which is nice!
Thank you to both of you!

One small remark perhaps: I noticed `self._rows = 500` but this default value is never used. If we don't call `set_rows()` then `self._default_args["rows"] = self._rows` is not called and thus no "rows" parameter is sent when launching the search job and so the server-side API's default is used :)

abowersox-cb:
Looks like we need to address the unit test coverage on this PR.

src/cbc_sdk/endpoint_standard/base.py
mrangelov-cb: Fixed.

cnotin: Confirmed :)
Default is now 500 rows in total, and no regression when using "set_rows()" 👍

PR #61, opened 2020-10-30
CBAPI-2018 Process and Process Event Query Methods

llyon-cb:
## Pull request checklist

https://jira.carbonblack.local/browse/CBAPI-2018

## Pull Request Description

Adds the ability to set criteria, exclusions, fields, sort, start, rows, and time_range for Process and Event searches
Moved the sort method up from the AsyncProcessQuery object into Query

- [ ] Yes
- [x] No

## How Has This Been Tested?

Pytests added


llyon-cb:
Documentation updates needed for RTD, specifically the Criteria section of concepts.rst

src/cbc_sdk/enterprise_edr/base.py
```diff
@@ -272,38 +270,189 @@ def __init__(self, doc_class, cb):
...
+        Updates a list of exclusion being collected for a query, by setting or appending items.
+
+        Args:
+            key (str): The key for the exclusion item to be set.
+            newlist (list): List of values to be set for the exclusion item.
+        """
+        oldlist = self._exclusions.get(key, [])
+        self._exclusions[key] = oldlist + newlist
+
+    def update_fields(self, fields):
```
avanbrunt-cb, line 341: It seems odd to have these (`update_fields`, `update_start`, `update_rows`, `update_time_range`) called update as they are more setters unlike criteria and exclusions where we are adding additional key/values to the request.

I would either go with `cb.select(Process).set_rows()` or just `cb.select(Process).rows()`

Now that I am thinking about this would it make sense to make `update_criteria` --> `add_criteria`?

avanbrunt-cb: This would also match closer to `sort_by` function naming

src/cbc_sdk/enterprise_edr/base.py
```diff
@@ -338,14 +487,15 @@ def _validate(self, args):
...
+        # iterate over total result set, 500 at a time
         args = self._get_query_parameters()
         self._validate(args)
 
-        if start != 0:
+        if start:
             args['start'] = start
-        args['rows'] = self._batch_size
+        if rows:
+            args['rows'] = rows
```
avanbrunt-cb, line 498: Is this suppose to be `if start is None:` and the same for rows?

llyon-cb: This is intended to be a check for if the params were specified. If they were, we add those keys and vals to our request. If not, we omit them and use the default backend values

src/cbc_sdk/enterprise_edr/base.py
llyon-cb: I think the methods for time_range, fields, start, and rows would make more sense as `set_*()` or `add_*()` since they're adding or overwriting the existing values. The criteria and exclusions methods are adding or updating the existing values, so those make more sense as `update_*()` or `add_*()`.

src/cbc_sdk/enterprise_edr/base.py
avanbrunt-cb: hmm. Then should the defaults for those not be None? `start=0, rows=20`? Looking through the code it looks like this would fail if not value was set for start or rows since you wouldn't be able to increment or check completion against start and rows

llyon-cb: Ahh yes that totally makes sense, I hadn't considered that. Good catch!

src/cbc_sdk/enterprise_edr/base.py
avanbrunt-cb: I wouldn't do `add_rows` as that reads a little strange `set_rows` makes more sense. Then I would go with `add_criteria` as update would make more sense if we were updating an existing criteria property like device_ids that is already set

src/cbc_sdk/enterprise_edr/base.py
```diff
@@ -280,25 +280,30 @@ def __init__(self, doc_class, cb):
...
             key (str): The key for the criteria item to be set.
-            newlist (list): List of values to be set for the criteria item.
+            newlist (str or list[str]): Value or list of values to be set for the criteria item.
 
         Returns:
             The ResultQuery with specified custom criteria.
 
         Example:
             query = api.select(Event).update_criteria("event_type", ["filemod", "scriptload"])
+            query = api.select(Event).update_criteria("event_type", "filemod")
```
avanbrunt-cb, line 295: function name example needs updated for both criteria and exclusions

avanbrunt-cb (approved):
LGTM. Two minor changes to the examples

abowersox-cb (approved):

PR #66, opened 2020-11-05
CBAPI-2058: add asynchronous query support to process and process event queries

abowersox-cb:
## Pull request checklist

 https://jira.carbonblack.local/browse/CBAPI-2058

## Pull Request Description

Add the asynchronous query functionality (execute_async()) to AsyncProcessQuery (the query for the Process object) and Query (the query for the Event object).

- [ ] Yes
- [x] No

## How Has This Been Tested?

New unit tests have been added to show that the asynchronous queries performs in the same manner as the synchronous queries.

### DIFF
--- a/src/tests/unit/enterprise_edr/test_enterprise_edr_events.py
+++ b/src/tests/unit/enterprise_edr/test_enterprise_edr_events.py
@@ -87,3 +87,22 @@ def test_event_query_select_with_where(cbcsdk_mock):
     assert len(results) == 100
     first_result = results[0]
     assert first_result.process_guid == guid
+
+
+def test_event_query_select_asynchronous(cbcsdk_mock):
+    """Test Event Querying with where() clause as asynchronous"""
+    search_validate_url = "/api/investigate/v1/orgs/test/events/search_validation"
+    cbcsdk_mock.mock_request("GET", search_validate_url, EVENT_SEARCH_VALIDATION_RESP)
+
+    url = r"/api/investigate/v2/orgs/test/events/J7G6DTLN\\-006633e3\\-00000334\\-00000000\\-1d677bedfbb1c2e/_search"
+    cbcsdk_mock.mock_request("POST", url, EVENT_SEARCH_RESP)
+
+    api = cbcsdk_mock.api
+    guid = "J7G6DTLN-006633e3-00000334-00000000-1d677bedfbb1c2e"
+
+    events = api.select(Event).where(process_guid=guid)
+    future = events.execute_async()
+    results = future.result()
+    assert len(results) == 250  # we get all of them when we run in background
+    first_event = results[0]
+    assert first_event['process_guid'] == guid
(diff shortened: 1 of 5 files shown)


src/tests/unit/enterprise_edr/test_enterprise_edr_query.py
```diff
@@ -160,3 +160,31 @@ def test_async_submit(cbcsdk_mock):
...
+                                     "2c292717-80ed-4f0d-845f-779e09470920"),
+                             GET_PROCESS_SEARCH_JOB_RESP)
+    # mock the GET to get search results
+    cbcsdk_mock.mock_request("GET", ("/api/investigate/v2/orgs/test/processes/search_jobs/"
+                                     "2c292717-80ed-4f0d-845f-779e09470920/results"),
+                             get_process_search_response)
+    process_query = api.select(Process).where(f"process_guid:{guid}")
+    future = process_query.execute_async()
+    results = future.result()
+    assert len(results) == 1
```
avanbrunt-cb, line 190: Can we add more validation beyond the length being 1? At the minimum verify the id of the process

avanbrunt-cb (approved):
One minor nitpick but otherwise looks good

src/tests/unit/enterprise_edr/test_enterprise_edr_query.py
abowersox-cb: Yes, except it'll add some complexity, given that the process_guid fields of the returned test data aren't (always) consistent with the guid parameters as given. But I can make that change if it's important.

avanbrunt-cb: It doesn't have to be the id. I'd just like to see one more level of assertion beyond the length to verify we are getting what we expect back

avanbrunt-cb: Perfect thank you!

abowersox-cb: OK, you have it, in the form of the process_pid. I modified the return data so that the pid would be different for the two cases, and added an extra assert to verify it against a new parameter.

src/tests/unit/enterprise_edr/test_enterprise_edr_events.py
```diff
@@ -87,3 +87,22 @@ def test_event_query_select_with_where(cbcsdk_mock):
...
+    url = r"/api/investigate/v2/orgs/test/events/J7G6DTLN\\-006633e3\\-00000334\\-00000000\\-1d677bedfbb1c2e/_search"
+    cbcsdk_mock.mock_request("POST", url, EVENT_SEARCH_RESP)
+
+    api = cbcsdk_mock.api
+    guid = "J7G6DTLN-006633e3-00000334-00000000-1d677bedfbb1c2e"
+
+    events = api.select(Event).where(process_guid=guid)
+    future = events.execute_async()
+    results = future.result()
+    assert len(results) == 250  # we get all of them when we run in background
```
avanbrunt-cb, line 106: This is only partially true since events works slightly different that the other search APIs. We need to check if `processed_segments == total_segements` 

Here is a link to the EA which required us to expose this property so they could retry the request to make sure they get the entire results back https://jira.carbonblack.local/browse/EA-16415?focusedCommentId=1230189&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-1230189

avanbrunt-cb (approved):
Looks good to me. Any way to recover some of the lost test coverage?

abowersox-cb:
> Looks good to me. Any way to recover some of the lost test coverage?

My latest change does that, because the base Query `_search()` method was no longer being called.

llyon-cb (approved):
LGTM

PR #67, opened 2020-11-09
Pip modules normalized names use dash instead of underscore

cnotin:
## Pull request checklist

It does not generate an issue correctly because the normalization is automatically applied by "pip" and "pypi"

## Pull Request Description
carbon_black_cloud_sdk -> carbon-black-cloud-sdk
See https://www.python.org/dev/peps/pep-0503/#normalized-names and https://pypi.org/project/carbon-black-cloud-sdk/

- [ ] Yes
- [x] No

## How Has This Been Tested?

I did not test the change in setup.py, and obviously I cannot submit to pypi so I let you check if it runs fine :)

### DIFF
--- a/docs/getting-started.rst
+++ b/docs/getting-started.rst
@@ -7,8 +7,8 @@ authentication for it, and executing a simple example program that makes one API
 
 Installation
 ------------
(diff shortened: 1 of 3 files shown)


avanbrunt-cb (approved):
Thanks and good catch!

llyon-cb (approved):

Issue #68, opened 2020-11-09
[REQUEST] PSC Alerts API: use larger "rows" value by default


cnotin:
This bug is a copy of https://github.com/carbonblack/cbapi-python/issues/275

**I am seeing this behaviour on: (please complete the following information):**
- PSC

**Describe the bug**
I use the PSC Alerts API and getting many alerts is slow...

**Steps to Reproduce**
1. Run an alerts search that yield around 1k alerts
2. It takes around 1 minute (contrary to 1 second if we request 1000 rows directly from the API)

**Expected behavior**
The current implementation doesn't set the "rows" parameter and thus currently the API returns pages of 20 results. Iterating on 20 rows chunks is very slow contrary to 50 (default in UI) or more, this is similar to https://github.com/carbonblack/cbapi-python/pull/252/.

Moreover this Python API doesn't offer to change this value.

PR #69, opened 2020-11-09
PSC Alerts API: requests pages of 100 rows by default

cnotin:
This is a duplicate of https://github.com/carbonblack/cbapi-python/pull/276
## Pull request checklist

- [x] Code follows the style guidelines of this project (PEP8, clean code).
- [ ] Linter has passed locally and any fixes were made for failures.
It's a small fix so I didn't do everything, please let me know if it's blocking...

## Pull request type
Please check the type of change your PR introduces:
- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes (not tied to bugs/features)
- [ ] Other (please describe):

- Issue Number: #68

- [ ] Yes
- [x] No

## How Has This Been Tested?

I ran a request with https://github.com/carbonblack/cbapi-python/blob/master/examples/psc/list_cbanalytics_alerts.py (adapted for this new library of course) that returns ~1k alerts. It was much faster and in my proxy I can see the additional `"rows": 100` parameter while the results are now in chunks of 100

## Other information:

Similar to https://github.com/carbonblack/cbapi-python/pull/252

I chose "100" arbitrarily but you might prefer a different default (by default, API returns 20 rows, while UI requests 50).
Moreover, this API client could also allow to pass this as a parameter (isn't implemented here)

### DIFF
--- a/src/cbc_sdk/platform/alerts.py
+++ b/src/cbc_sdk/platform/alerts.py
@@ -744,6 +744,7 @@ class BaseAlertSearchQuery(PlatformQueryBase, QueryBuilderSupportMixin, Iterable
         """
         request = {"criteria": self._build_criteria()}
         request["query"] = self._query_builder._collapse()
+        # Fetch 100 rows per page (instead of 10 by default) for better performance
         request["rows"] = 100
         if from_row > 0:
             request["start"] = from_row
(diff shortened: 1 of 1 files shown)


avanbrunt-cb (commented):
This is a good addition to provide context on the default though can you add a comment above that line indicating why we are setting the default to 100.

cnotin:
Here you go :)

avanbrunt-cb (approved):

llyon-cb (approved):

PR #72, opened 2020-11-12
CBAPI 2020 Process Facet Search

llyon-cb:
## Pull request checklist

https://jira.carbonblack.local/browse/CBAPI-2020
## Pull Request Description

Add Process Facet search

- [ ] Yes
- [x] No

## How Has This Been Tested?

Pytest


src/cbc_sdk/base.py
```diff
@@ -321,20 +321,41 @@ def __str__(self):
...
-                val = repr(self._info[attr])
-            if len(val) > 50:
-                val = val[:47] + u"..."
-            lines.append(u"{0:s} {1:>20s}: {2:s}".format(status, attr, val))
+        try:
+            attributes = sorted(self._info)
+        except:
+            attributes = self._info
+
+        for attr in attributes:
```
avanbrunt-cb, line 329: Can we add some comments here to describe the modified behavior and what the two paths are

src/cbc_sdk/enterprise_edr/base.py
```diff
@@ -237,6 +246,103 @@ def children(self):
...
+            """Initialize a ProcessFacet Terms object with initial_data."""
+            super(ProcessFacet.Terms, self).__init__(
+                cb,
+                model_unique_id=None,
+                initial_data=initial_data,
+                force_init=False,
+                full_doc=True,
+            )
+            self._facets = {}
+            for facet_term_data in initial_data:
```
avanbrunt-cb, line 277: Is there a possibility that `initial_data` is None due to an uncaught error?

src/cbc_sdk/enterprise_edr/base.py
```diff
@@ -698,6 +804,271 @@ def _run_async_query(self, context):
...
+        super(AsyncFacetQuery, self).__init__(doc_class, cb)
+        self._query_token = None
+        self._timeout = 0
+        self._timed_out = False
+        self._limit = None
+        self._facet_fields = []
+        self._ranges = []
+        self._facet_rows = None
+
+    def timeout(self, msecs):
```
avanbrunt-cb, line 819: Can we mention the default in the docs here?

src/cbc_sdk/enterprise_edr/base.py
llyon-cb: I don't think so. It could be an empty list `[]`, but not None without a `ServerError` being raised. In `connection.BaseAPI`, the function `get_object` performs a GET on the job results URL with optional query params. If the response from CB isn't code 200, or it's unable to parse the response as JSON, a `ServerError` gets raised.

If the API returned 204, then initial_data could be None. I think a 204 would only come back on a DELETE, otherwise if there's no matching data to return, the API gives back "terms" and "ranges" as empty JSON objects. In that case, we'd have `initial_data == []`

avanbrunt-cb: Thanks for the explanation

avanbrunt-cb (approved):
LGTM

abowersox-cb (approved):
Good going, Luke!

mrangelov-cb (approved):

PR #73, opened 2020-11-16
CBPAPI-2026 Enriched Events Facet Search

mrangelov-cb:
## Pull request checklist

- [ ] Yes
- [x] No


src/cbc_sdk/endpoint_standard/base.py
```diff
@@ -616,3 +697,297 @@ def _search(self, start=0, rows=0):
...
+            raising an exception
+        """ 
+        self._query_builder.or_(None, **kwargs)
+        return self
+        
+
+    def _get_query_parameters(self):
+        """Need to override base class implementation as it sets custom (invalid) fields"""
+        args = self._default_args.copy()
+        if not self._facet_fields:
```
llyon-cb, line 731: I just found that it's OK to have `_facet_fields` undefined as long as `_ranges` is defined, so I would just switch this check to `if not (self._facet_fields or self._ranges):`

src/cbc_sdk/endpoint_standard/base.py
```diff
@@ -616,3 +697,297 @@ def _search(self, start=0, rows=0):
...
+            query_parameters = {}
+
+        result = self._cb.get_object(result_url, query_parameters=query_parameters)
+        result["job_id"] = self._query_token
+        yield result
+
+    def _search_async(self):
+        """ Not very pretty hack to incorporate both synchronous and async behavior """
+        for result in self._search():
+            result.pop("job_id", None)
```
llyon-cb, line 973: Is it necessary to remove "job_id" before creating the EnrichedEventFacet object from the result?

src/cbc_sdk/endpoint_standard/base.py
mrangelov-cb: Indeed. Changed.

src/cbc_sdk/endpoint_standard/base.py
mrangelov-cb: Works fine with job_id in result, so no need to remove. Fixed.

src/cbc_sdk/endpoint_standard/base.py
```diff
@@ -616,3 +697,297 @@ def _search(self, start=0, rows=0):
 
             log.debug("current: {}, total_results: {}".format(current, self._total_results))
 
+
+class EnrichedEventFacetQuery(Query, AsyncQueryMixin):
```
llyon-cb, line 701: This class could also use methods for adding criteria and exclusions. Platform Queries have methods for these that could be copied into this Query Class https://github.com/carbonblack/carbon-black-cloud-sdk-python/blob/ec2bb7d16d61755c71e30653c3a813ccf0f76460/src/cbc_sdk/enterprise_edr/base.py#L390

src/cbc_sdk/base.py
```diff
@@ -321,21 +321,42 @@ def __str__(self):
...
-                    status = "(*)"
-            try:
-                val = str(self._info[attr])
-            except UnicodeDecodeError:
-                val = repr(self._info[attr])
-            if len(val) > 50:
-                val = val[:47] + u"..."
-            lines.append(u"{0:s} {1:>20s}: {2:s}".format(status, attr, val))
-
+        try:
```
avanbrunt-cb, line 324: @llyon-cb @mrangelov-cb We can't add this code in both PRs otherwise we will get a conflict on merge. One of these PRs will need to manage this conflict after one of the PRs is merged. Please discuss who is going to fix the conflict and if any help is needed

src/cbc_sdk/endpoint_standard/models/enrichedEventFacet.yaml
```diff
@@ -0,0 +1,68 @@
+type: object
```
avanbrunt-cb, line 1: This file should be snake_case. The other files in this models folder are old and don't use the new style

src/tests/unit/endpoint_standard/test_endpoint_standard_enriched_events_facet.py
```diff
@@ -0,0 +1,201 @@
...
+def test_enriched_event_facet_select_where(cbcsdk_mock):
+    """Testing EnrichedEvent Querying with select()"""
+    cbcsdk_mock.mock_request("POST", "/api/investigate/v2/orgs/test/enriched_events/facet_jobs", POST_ENRICHED_EVENTS_FACET_SEARCH_JOB_RESP)
+    cbcsdk_mock.mock_request("GET", "/api/investigate/v2/orgs/test/enriched_events/facet_jobs/08ffa932-b633-4107-ba56-8741e929e48b/results", GET_ENRICHED_EVENTS_FACET_SEARCH_JOB_RESULTS_RESP_1)
+
+    api = cbcsdk_mock.api
+    events = api.select(EnrichedEventFacet).where(process_name="chrome.exe").add_facet_field("process_name")
+    for event in events:
+        assert event.terms is not None
+        assert event.ranges is not None
```
avanbrunt-cb, line 43: This needs to a more detailed assert than not None as event.ranges was not request only facet field. This should be verified as an empty array and the terms should only contain the process_name term

avanbrunt-cb: Please apply this to all the tests added in this PR

src/tests/unit/endpoint_standard/test_endpoint_standard_enriched_events_facet.py
```diff
@@ -0,0 +1,201 @@
...
+
+@pytest.fixture(scope="function")
+def cbcsdk_mock(monkeypatch, cb):
+    """Mocks CBC SDK for unit tests"""
+    return CBCSDKMock(monkeypatch, cb)
+
+
+# ==================================== UNIT TESTS BELOW ====================================
+
+def test_enriched_event_facet_select_where(cbcsdk_mock):
```
avanbrunt-cb, line 34: There needs to be a test for using add_facet_field([str]) we should probably have a test for an unexpected value like a number or None as well

src/cbc_sdk/base.py
mrangelov-cb: I reverted the change, as it doesn't break any tests.

src/cbc_sdk/endpoint_standard/base.py
mrangelov-cb: I noticed those are implemented in your FacetQuery base class, I guess inheriting from it once it gets merged would by itself solve that problem?

src/cbc_sdk/endpoint_standard/models/enrichedEventFacet.yaml
mrangelov-cb: Done.

src/tests/unit/endpoint_standard/test_endpoint_standard_enriched_events_facet.py
mrangelov-cb: Done.

src/tests/unit/endpoint_standard/test_endpoint_standard_enriched_events_facet.py
mrangelov-cb: Done.

avanbrunt-cb (approved):
LGTM

abowersox-cb (approved):

PR #74, opened 2020-11-17
CBAPI-2019: added device processes example script

abowersox-cb:
## Pull request checklist

 https://jira.carbonblack.local/browse/CBAPI-2019

## Pull Request Description

Added an example script to show multiple process searches running asynchronously, one for each device.

- [ ] Yes
- [x] No

## How Has This Been Tested?

Script has been run against prod02.

### DIFF
--- a/examples/enterprise_edr/device_processes.py
+++ b/examples/enterprise_edr/device_processes.py
@@ -1,0 +1,52 @@
+#!/usr/bin/env python
+# *******************************************************
+# Copyright (c) VMware, Inc. 2020. All Rights Reserved.
+# SPDX-License-Identifier: MIT
+# *******************************************************
+# *
+# * DISCLAIMER. THIS PROGRAM IS PROVIDED TO YOU "AS IS" WITHOUT
+# * WARRANTIES OR CONDITIONS OF ANY KIND, WHETHER ORAL OR WRITTEN,
+# * EXPRESS OR IMPLIED. THE AUTHOR SPECIFICALLY DISCLAIMS ANY IMPLIED
+# * WARRANTIES OR CONDITIONS OF MERCHANTABILITY, SATISFACTORY QUALITY,
+# * NON-INFRINGEMENT AND FITNESS FOR A PARTICULAR PURPOSE.
+
+import sys
+import logging
+import concurrent.futures
+from cbc_sdk.helpers import eprint, build_cli_parser, get_cb_cloud_object
+from cbc_sdk.platform import Device
+from cbc_sdk.enterprise_edr import Process
+
+log = logging.getLogger(__name__)
+
+
+def main():
+    parser = build_cli_parser()
+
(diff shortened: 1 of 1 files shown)


examples/enterprise_edr/device_processes.py
```diff
@@ -0,0 +1,52 @@
...
+import logging
+import concurrent.futures
+from cbc_sdk.helpers import eprint, build_cli_parser, get_cb_cloud_object
+from cbc_sdk.platform import Device
+from cbc_sdk.enterprise_edr import Process
+
+log = logging.getLogger(__name__)
+
+
+def main():
```
avanbrunt-cb, line 23: Can we add some comments around the usage pattern for the cli input

examples/enterprise_edr/device_processes.py
```diff
@@ -0,0 +1,52 @@
...
+    parser = build_cli_parser()
+
+    args = parser.parse_args()
+    cb = get_cb_cloud_object(args)
+
+    query = cb.select(Device).where('1')
```
avanbrunt-cb, line 29: Let's make this a query for active devices using "ACTIVE" for device_status

avanbrunt-cb (approved):
Looks great only two small pieces of feedback

examples/enterprise_edr/device_processes.py
abowersox-cb: How do you write that? Is it just "device_status:ACTIVE"?

examples/enterprise_edr/device_processes.py
abowersox-cb: Why? None of our other examples has those.

avanbrunt-cb: This script does not have any descriptions of what it does we need at a minimum some context for what this does so a developer can't better utilize the example. Our other examples should have descriptions too

examples/enterprise_edr/device_processes.py
avanbrunt-cb: It's a criteria property https://developer.carbonblack.com/reference/carbon-black-cloud/platform/latest/devices-api/

avanbrunt-cb: https://github.com/carbonblack/carbon-black-cloud-sdk-python/blob/develop/src/cbc_sdk/platform/devices.py#L335

abowersox-cb: OK, I remember now.  Adding the set_status() to the query reduced the number of devices being queried from 116 to 15.

examples/enterprise_edr/device_processes.py
abowersox-cb: I added a docstring to the script to explain what exactly it does.  I suggest that an additional story be created to add those to other examples.

llyon-cb (approved):

PR #76, opened 2020-11-18
[WIP] Update EnrichedEventFacet to use base.FacetQuery

llyon-cb:
Update EnrichedEventFacet to use FacetQuery; update tests


llyon-cb:
Hey Milen, I've tried to update your branch to use the new `cbc_sdk.base.FacetQuery` class. The biggest change is how you retrieve results. Instead of creating a list from the generator (i.e. `facets = [facet for facet in event_facet_query]`), we now use the `.results` property (i.e. facets = event_facet_query.results`). 

The change in getting results is mainly because we only ever expect one object to be returned from the Facet API's (containing the terms and ranges objects). 

Since the two facet query classes share so much code, it made sense to combine them to hopefully reduce complexity. LMK if you have suggestions for how this could be done better.

I mentioned in a slack DM that when these queries are executed synchronously, they ended up using methods that were written for pagination and iteration, which made them messy to override. This seems like a good solution to that problem.

mrangelov-cb:
That looks great. I started doing similar changes myself when I saw what you posted on slack, however it looks like the FacetQuery code is not merged in develop yet, so I decided to hold on until your process facet search PR gets merged first. Looks like it works slightly different to how I imagined.  Thanks a lot for this!

mrangelov-cb (approved):

PR #78, opened 2020-11-30
RTD update, initial commit

mrangelov-cb:
## Pull request checklist

- [ ] Yes
- [x] No


docs/concepts.rst
```diff
@@ -271,3 +271,89 @@ Modules not yet Supported for Criteria
...
+
+A number of queries allow for asynchronous mode of operation. Those utilize python futures and the request itself is performed in a separate worker thread.
+An internal thread pool is utilized to support multiple CBC queries executing in an asynchronous manner without blocking the main thread.
+
+Execute an asynchronous query
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Running asynchronous queries is done by invoking the ``execute_async()`` method, e.g:
+
+  >>> async_query = api.select(EnrichedEvent).where('process_name:chrome.exe')
```
avanbrunt-cb, line 287: You didn't include the execute_async() in the example

docs/concepts.rst
```diff
@@ -271,3 +271,89 @@ Modules not yet Supported for Criteria
...
+
+Facet search queries return statistical information indicating the relative weighting of the requested values as per the specified criteria.
+There are two types of criteria that can be set, one is the ``range`` type which is used to specify discrete values (integers or timestamps - specified both as seconds since epoch and also as ISO 8601 strings).
+The results are then grouped by occurence within the specified range.
+The other type is the ``term`` type which allow for one or more fields to use as a criteria on which to return weighted results.
+
+Setting ranges
+^^^^^^^^^^^^^^
+
+Ranges are configured via the ``add_range()`` method which accepts a dictionary of range settings:
```
avanbrunt-cb, line 321: Out of curiosity why are we passing ranges as a dictionary instead of requiring the four properties in the function call? Using a dictionary makes it less intuitive when you look at the function call. Are we gaining a benefit somewhere?

docs/concepts.rst
mrangelov-cb: Oops...fixed

docs/concepts.rst
mrangelov-cb: This comes from the base class, FacetQuery, the reason being we can add either a single range or a list of ranges in one call as opposed to separate calls to add individual ranges. We do the same for terms. It would be problematic to add a list of criteria as function arguments. I think this might need a discussion (ATM both PRs aren't merged yet).

avanbrunt-cb: That's not how this comes across though it looks like it only accepts a dictionary not an array of dictionaries

docs/concepts.rst
```diff
@@ -271,3 +271,91 @@ Modules not yet Supported for Criteria
 :mod:`RunHistory <cbc_sdk.audit_remediation.base.RunHistory>`
 :mod:`Event <cbc_sdk.enterprise_edr.base.Event>`
 :mod:`Process <cbc_sdk.enterprise_edr.base.Process>`
```
llyon-cb, line 273: With our most recent updates, the Event and Process modules now support criteria

docs/concepts.rst
```diff
@@ -271,3 +271,91 @@ Modules not yet Supported for Criteria
...
+Results from asynchronous queries can be retrieved by using the result() method since they are actually futures:
+
+  >>> print(async_query.result())
+
+This would block the main thread until the query completes.
+
+Modules with support for asynchronous queries
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+:mod:`Process <cbc_sdk.enterprise_edr.base.Process>`
```
llyon-cb, line 303: Could we add Process Facets to this list as well?

`:mod:ProcessFacet <cbc_sdk.enterprise_edr.base.ProcessFacet>`

docs/concepts.rst
mrangelov-cb: Got it. Corrected.

docs/concepts.rst
mrangelov-cb: Added them.

docs/concepts.rst
mrangelov-cb: Added that as well

llyon-cb (approved):

avanbrunt-cb (approved):

docs/concepts.rst
```diff
@@ -265,9 +265,101 @@ Modules with Support for Criteria
...
+* ``bucket_size``- how large of a bucket to group results in. If grouping an ISO 8601 property, use a string like '-3DAYS'
+
+Multiple ranges can be configured per query by passing a list of range dictionaries.
+
+Setting terms
+^^^^^^^^^^^^^
+
+Terms are configured via the ``add_facet_field()`` method:
+
+    >>> query = api.select(EnrichedEventFacet).where(process_pid=1000),add_facet_field("process_name")
```
kebringer-cb, line 349: There's a comma before the add_facet which should be a period I think.

mrangelov-cb: Fixed.

PR #85, opened 2020-12-08
Enriched Event Details support

mrangelov-cb:
## Pull request checklist

- [ ] Yes
- [x] No


src/cbc_sdk/endpoint_standard/base.py
```diff
@@ -617,6 +620,15 @@ def timeout(self, msecs):
         self._timeout = msecs
         return self
 
+    def get_details(self):
```
avanbrunt-cb, line 623: This should not be on the EnrichedEventQuery instead this should be on the EnrichedEvent model

avanbrunt-cb: The current implementation will overload the Details API when an EnrichedEventQuery matches more than 100 events. The best practice indicated by the Nic Cage team also recommends we fetch the details for a singular event as that is more optimal for the backend.

100 limit is documented https://developer.carbonblack.com/reference/carbon-black-cloud/platform/latest/platform-search-api-enriched-events/#request-details-for-enriched-events-v2

avanbrunt-cb (changes requested):
The get_details function must be on the EnrichedEvent model not the query

src/cbc_sdk/endpoint_standard/base.py
mrangelov-cb: Done.

src/cbc_sdk/endpoint_standard/base.py
```diff
@@ -258,9 +258,77 @@ def __init__(self, cb, model_unique_id=None, initial_data=None, force_init=False
...
+            self._cb.credentials.org_key,
+            job_id
+        )
+        query_parameters = {}
+        while still_fetching:
+            result = self._cb.get_object(result_url, query_parameters=query_parameters)
+            total_results = result.get('num_available', 0)
+            if total_results != 0:
+                results = result.get('results', [])
+                return EnrichedEvent(self._cb, initial_data=results[0])
```
avanbrunt-cb, line 330: Why return a new model instead of modifying the existing model and returning self?

avanbrunt-cb:
```
self._info = results[0]
return self
```

avanbrunt-cb (approved):
One last suggestion so you can merge if you modify the code

llyon-cb (approved):

src/cbc_sdk/endpoint_standard/base.py
mrangelov-cb: That was actually what bothered me about this...it felt as if I should not return a modified self, but rather the details of a search, then again when I return a new model instane, you could call get_details() on it and it makes no sense. I don't really know what would be best in that case :(

avanbrunt-cb: So in essence you are not modifying the object but rather enhancing it with additional details. The details search only fetches additional fields and any that might have changed since the original time the event was fetched. You have a point about having already fetched the details do you want to block a second fetch. I'm not opposed but then again that wouldn't be common and it's not the end of the world so it might be better left alone.

mrangelov-cb: OK, modifying the object rather than instantiating a new one now.

Issue #83, opened 2020-12-07
[BUG] Process.events() where processed_segments != total_segments may result in infinite loop


csanders-git:
**I am seeing this behaviour on: (please complete the following information):**
Python 3.8, latest Master branch

**Describe the bug**

- It seems that there is some logic provided to deal with the not well defined concept of segments within the API. [1]
- We have been experiencing an issue with the API whereby we've entered into what appears to be an infinite loop. 
- We have a response where _processed_segments != _total_segments (derived from Process.events).
- In this condition the wrapper will continue to request this Event information in a `while` loop until `_processed_segments` == `_total_segments` 
- We have observed that the wrapper has been re-requesting the EventQuery POST request [2] for the past week without change to the process event data response.

[1] (https://github.com/carbonblack/carbon-black-cloud-sdk-python/blob/d65c31a1ad2d64677bfc169e33687faf2b68e054/src/cbc_sdk/enterprise_edr/base.py#L969)

[2](https://github.com/carbonblack/carbon-black-cloud-sdk-python/blob/d65c31a1ad2d64677bfc169e33687faf2b68e054/src/cbc_sdk/enterprise_edr/base.py#L962). 

While debugging we continuously see 
```
{'results': [...], 'num_found': 1, 'num_available': 1, 'total_segments': 5833, 'processed_segments': 5501}
```

**Steps to Reproduce**
Its somewhat unclear what how to reproduce this as processed_segments and total_segments don't have much documentation as to their purpose within official docs (https://developer.carbonblack.com/reference/carbon-black-cloud/platform/latest/platform-search-api-processes/). 

What we do observe is that this shortcoming seems to be a result of either a failure on implementation of the backend or a misjudgement by the API wrapper on how long this reconciliation will take. 

**Expected behavior**
We'd expect that if these responses may take seemingly infinite (or longer than the expected length of most application runs) amount of time to finish that there be a max loop iteration for these enrichments.

abowersox-cb:
Pull request #89 adds a retry limit to the event query loops.  The limit is 10 retries, but the counter is reset by advancement of the "processed_segments" count, as long as there is forward progress.  @csanders-git please check out the latest from the develop branch and verify the fix.

avanbrunt-cb:
This has been fixed with the latest release

PR #89, opened 2020-12-10
CBAPI-2183: added a limit to the retry mechanism on event & event facet queries

abowersox-cb:
## Pull request checklist

https://jira.carbonblack.local/browse/CBAPI-2183 (issue #83)

## Pull Request Description

Add a mechanism to keep from infinitely looping in event and event facet queries where the returned "processed_segments" never reaches the returned "total_segments." The mechanism uses a retry counter that is not incremented (and is, in fact, reset) as long as the "processed_segments" counter is making forward progress; only if it "stalls" does our retry counter get incremented, and, at 10 retries, a TimeoutError is thrown.

- [ ] Yes
- [x] No

(Semi-breaking change: Event and event facet queries may now throw TimeoutError in addition to ApiError.  This only affects code that wants to distinguish the two cases.)

## How Has This Been Tested?

Additional unit tests have been added that trigger the TimeoutError mechanism.

### DIFF
--- a/src/cbc_sdk/platform/base.py
+++ b/src/cbc_sdk/platform/base.py
@@ -1015,9 +1015,10 @@ class EventQuery(Query):
             self._count_valid = True
             if self._processed_segments != self._total_segments:
                 retry_counter = 0 if self._processed_segments > last_processed_segments else retry_counter + 1
-                last_processed_segments = self._processed_segments
+                last_processed_segments = max(last_processed_segments, self._processed_segments)
                 if retry_counter == MAX_EVENT_SEARCH_RETRIES:
                     raise TimeoutError(url, resp.status_code, "excessive number of retries in event query")
+                time.sleep(1 + retry_counter / 10)
                 continue  # loop until we get all segments back
 
             results = result.get('results', [])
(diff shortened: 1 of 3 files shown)


src/cbc_sdk/platform/base.py
```diff
@@ -957,6 +959,10 @@ def _search(self, start=0, rows=0):
             self._processed_segments = result.get("processed_segments", 0)
             self._count_valid = True
             if self._processed_segments != self._total_segments:
+                retry_counter = 0 if self._processed_segments > last_processed_segments else retry_counter + 1
```
avanbrunt-cb, line 962: There probably needs to be some level of delay/sleep otherwise the 10 iterations will go too quickly. This does need to take at a minimum 10-15 seconds

abowersox-cb: Addressed.

avanbrunt-cb (approved):

src/cbc_sdk/platform/base.py
```diff
@@ -957,6 +959,11 @@ def _search(self, start=0, rows=0):
...
+                last_processed_segments = max(last_processed_segments, self._processed_segments)
+                if retry_counter == MAX_EVENT_SEARCH_RETRIES:
+                    raise TimeoutError(url, resp.status_code, "excessive number of retries in event query")
+                time.sleep(1 + retry_counter / 10)
```
llyon-cb, line 966: Any particular reason you chose a divisor of 10? 100ms doesn't seem like an appreciable difference in sleep times to me

llyon-cb (approved):
Just one quick question, otherwise looks good!

src/cbc_sdk/platform/base.py
abowersox-cb: Alex mentioned that the overall wait should be 10-15 seconds.  This was an easy way to add a slight bit of backoff to the individual wait cycles; total wait would then be 14.5 seconds.

PR #113, opened 2021-01-26
Add set_time_range to alerts. Move add/update_criteria to Mixin

avanbrunt-cb:
## Pull request checklist

Slack Thread request

## Pull Request Description

* Add support for additional criteria time ranges for Alerts
* Up level _update_criteria and update_criteria functions to a mixin to be added easily to any model that needs criteria
* Up level add_criteria to mixin and remove duplicate _add_criteria as this replicates _update_criteria

- [ ] Yes
- [x] No

## How Has This Been Tested?

Additional unit tests and existing tests pass

### DIFF
--- a/src/cbc_sdk/base.py
+++ b/src/cbc_sdk/base.py
@@ -1428,6 +1428,41 @@ class QueryBuilderSupportMixin:
         return self
 
 
+class CriteriaBuilderSupportMixin:
+    """A mixin that supplies wrapper methods to access the _crtieria."""
+
+    def update_criteria(self, key, newlist):
+        """Update the criteria on this query with a custom criteria key.
+
+        Args:
+            key (str): The key for the criteria item to be set.
+            newlist (list): List of values to be set for the criteria item.
+
+        Returns:
+            The query object with specified custom criteria.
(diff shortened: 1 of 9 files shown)


src/cbc_sdk/base.py
```diff
@@ -1428,6 +1428,73 @@ def not_(self, q=None, **kwargs):
         return self
 
 
+class CriteriaBuilderSupportMixin:
```
abowersox-cb, line 1431: This makes the "criteria" support somewhat more complex, but I'm sure the complexity is in a good cause.

abowersox-cb (approved):
Tricky but I see what you're driving at.

src/cbc_sdk/base.py
avanbrunt-cb: This is simpler instead of having _update_criteria, update_criteria, and add_criteria in every query implementation that implements the functions the exact same way. Though this I also noticed the difference in behavior between alerts, device, and few other models and processes/events so I made the experience consistent across all models now.

emitreva-cb (approved):

PR #126, opened 2021-02-03
CBAPI-2267: Workloads - SDK Code changes - Vulnerability Assessment

emitreva-cb:
## Pull request checklist

https://jira.carbonblack.local/browse/CBAPI-2267

## Pull Request Description
- First portion of the APIs covering the Organization Level Vulnerabilities

- [ ] Yes
- [x] No


abowersox-cb:
Question from the floor: Should all of the Workload API code actually go under "platform," or, since we seem to be organizing workload APIs as their own category in the Dev Network documentation, go under a new top-level directory named "workload"?  This is relevant to my interests.

avanbrunt-cb:
> Question from the floor: Should all of the Workload API code actually go under "platform," or, since we seem to be organizing workload APIs as their own category in the Dev Network documentation, go under a new top-level directory named "workload"? This is relevant to my interests.

That's a great point we should probably make a workloads package then

abowersox-cb:
> That's a great point we should probably make a workloads package then

I would use "workload" (singular), following the name as given in the Dev Network documentation.

emitreva-cb:
> > That's a great point we should probably make a workloads package then
> 
> I would use "workload" (singular), following the name as given in the Dev Network documentation.

So should I move it directly under cbc_sdk?

abowersox-cb:
> So should I move it directly under cbc_sdk?

Create a directory named "workload" under cbc_sdk and move it there.

src/cbc_sdk/workload/vulnerability_assessment.py
```diff
@@ -0,0 +1,902 @@
...
+            the server.
+            **kwargs (dict): Not used, retained for compatibility.
+
+        Returns:
+            VulnerabilityQuery: The query object
+        """
+        return VulnerabilityQuery(cls, cb)
+
+    @classmethod
+    def get_vulnerability_summary(cls, cb, severity=None, vcenter_id=None):
```
avanbrunt-cb, line 94: This should include organization in the function name. maybe `get_org_vulnerability_summary`

src/cbc_sdk/workload/vulnerability_assessment.py
```diff
@@ -0,0 +1,902 @@
...
+            str: The URL used to make requests for this object.
+        """
+        return self.urlobject_single.format(self._model_unique_id)
+
+    def affected_assets(self, os_product_id=None):
+        """Returns a list of Vulnerability associated with device.
+
+        Returns:
+            Vulnerability ([Vulnerability]): List of Vulnerability in this device.
+            The os_product_id is required parameter in the request
```
avanbrunt-cb, line 176: I don't think this docstring matches the functionality of this API as you should not default the os_product_id to None and this does not return a List of Vulnerability in this device but a AffectedAssetQuery

lhilmes-cb (approved):
Needs a more "technical" set of eyes, but looks good to me!

src/cbc_sdk/workload/vulnerability_assessment.py
```diff
@@ -0,0 +1,902 @@
...
+        """
+        if os_product_id:
+            self._os_product_id = os_product_id
+            return AffectedAssetQuery(self, self._cb)
+        else:
+            raise ApiError("The os product id is required parameter")
+
+
+class DeviceVulnerability(NewBaseModel):
```
avanbrunt-cb, line 185: This class has no query implementation and returns the API results directly from get request. We often create objects with the results of the response ie
```
for x in results:
  yield Resource(cb, initial_data=x)
```

src/cbc_sdk/workload/vulnerability_assessment.py
```diff
@@ -0,0 +1,902 @@
...
+        Returns:
+            list: Result of the async query, which is then returned by the future.
+        """
+        url = self._build_url("/_search")
+        request = self._build_request(0, -1)
+        resp = self._cb.post_object(url, body=request)
+        result = resp.json()
+        self._total_results = result["num_found"]
+        self._count_valid = True
+        results = result.get("results", [])
```
avanbrunt-cb, line 882: See here as an example of how to return the correct results as a model https://github.com/carbonblack/carbon-black-cloud-sdk-python/blob/develop/src/cbc_sdk/platform/devices.py#L544

src/cbc_sdk/workload/vulnerability_assessment.py
```diff
@@ -0,0 +1,902 @@
...
+            request = self._build_request(current, max_rows)
+            resp = self._cb.post_object(url, body=request)
+            result = resp.json()
+
+            self._total_results = result["num_found"]
+            self._count_valid = True
+
+            results = result.get("results", [])
+            for item in results:
+                yield item
```
avanbrunt-cb, line 855: We should be building the DeviceVulnerability model similar to how we do it here https://github.com/carbonblack/carbon-black-cloud-sdk-python/blob/develop/src/cbc_sdk/platform/devices.py#L507

src/cbc_sdk/workload/vulnerability_assessment.py
```diff
@@ -0,0 +1,902 @@
...
+
+class DeviceVulnerabilityQuery(VulnerabilityQuery):
+    """Query Class for the DeviceVulnerability"""
+
+    def __init__(self, doc_class, cb):
+        """
+        Initialize the DeviceVulnerabilityQuery.
+
+        Args:
+            doc_class (class): The model class that will be returned by this query.
```
avanbrunt-cb, line 776: This needs to be renamed as you are not going to return a Device which is what you are receiving here instead you will be returning DeviceVulnerability. I think it's fine to hard code the `self.doc_class = DeviceVulnerability` so you should rename the parameter to device and store this into a separate variable `self.device` or something

src/cbc_sdk/workload/vulnerability_assessment.py
```diff
@@ -0,0 +1,902 @@
...
+
+    def __init__(self, doc_class, cb):
+        """
+        Initialize the VulnerabilityQuery.
+
+        Args:
+            doc_class (class): The model class that will be returned by this query.
+            cb (BaseAPI): Reference to API object used to communicate with the server.
+        """
+        self._doc_class = doc_class
```
avanbrunt-cb, line 638: This should be `self.vulnerability` instead of doc_class and `self.doc_class = Device`  see the other comment in DeviceVulnerability. Something that might be worth while to do is to gather all of the device_ids returned in the response and make a Device query like `self._cb.select(Device).set_device_ids([ids])` to return all the device models instead of just returning a list of devices

avanbrunt-cb (changes requested):

src/cbc_sdk/workload/vulnerability_assessment.py
```diff
@@ -0,0 +1,949 @@
...
+        Returns:
+            str: The URL used to make requests for this object.
+        """
+        return self.urlobject_single.format(self._model_unique_id)
+
+    def affected_assets(self, os_product_id):
+        """Returns a list of Vulnerability associated with device.
+
+        Args;
+            os_product_id (string) required parameter
```
avanbrunt-cb, line 179: The description reads a little strange replace it with `Returns a list of Vulnerability objects associated with device.` 

The Args is not quite right and needs a description
```
Args:
            os_product_id (str): {insert description}
```

src/cbc_sdk/workload/vulnerability_assessment.py
```diff
@@ -0,0 +1,949 @@
...
+            cb (BaseAPI): Reference to API object used to communicate with the server.
+            **kwargs (dict): Not used, retained for compatibility.
+
+        Returns:
+            DeviceVulnerabilityQuery: The query object
+        """
+        return DeviceVulnerabilityQuery(cls, cb)
+
+    @classmethod
+    def get_vulnerability_summary_per_device(cls, cb, device, category=None, vcenter_id=None):
```
avanbrunt-cb, line 239: This should be device_id instead of a device object. The doc_string is also missing some of the parameters.

avanbrunt-cb (commented):
Two more small changes. I didn't verify every function but make sure your docstrings match the args and returns

avanbrunt-cb (approved):
Going to approve to not block the PR next week so please verify all doc strings. The last two comments are small

src/tests/unit/workload/test_vulnerability_assessment.py
```diff
@@ -0,0 +1,451 @@
...
+    def post_validate(url, body, **kwargs):
+        assert body.get('os_product_id') is not None
+        return GET_AFFECTED_ASSETS_SPECIFIC_VULNERABILITY
+    cbcsdk_mock.mock_request("GET", "/vulnerability/assessment/api/v1/vulnerabilities/cveid",
+                             GET_VULNERABILITY_RESP)
+    cbcsdk_mock.mock_request("POST", "/appservices/v6/orgs/test/devices/_search", POST_DEVICE_SEARCH_RESP)
+    cbcsdk_mock.mock_request("POST",
+                             "/vulnerability/assessment/api/v1/orgs/test/vcenters/someid/vulnerabilities/cveid/devices",
+                             post_validate)
+    api = cbcsdk_mock.api
```
hkaragitliev-cb, line 312: accidental copy I guess

src/tests/unit/workload/test_vulnerability_assessment.py
```diff
@@ -0,0 +1,451 @@
...
+
+def test_get_affected_assets_fail(cbcsdk_mock):
+    """Test Get a list of assets affected by a specific vulnerability CVE ID."""
+    cbcsdk_mock.mock_request("GET", "/vulnerability/assessment/api/v1/vulnerabilities/CVE-2019-0708",
+                             GET_VULNERABILITY_RESP)
+    cbcsdk_mock.mock_request("POST",
+                             "/vulnerability/assessment/api/v1/orgs/test/vulnerabilities/CVE-2019-0708/devices",
+                             GET_AFFECTED_ASSETS_SPECIFIC_VULNERABILITY)
+    api = cbcsdk_mock.api
+    api = cbcsdk_mock.api
```
hkaragitliev-cb, line 329: duplicated line here as well

src/cbc_sdk/workload/vulnerability_assessment.py
abowersox-cb: Agreeing with Alex: use device_id (int) here instead of device (object).

hkaragitliev-cb (approved):

abowersox-cb (approved):
I'm good with this now.

mrangelov-cb (approved):

PR #128, opened 2021-02-05
CBAPI-2285: Workloads - SDK Code changes - Sensor Lifecycle Management

abowersox-cb:
## Pull request checklist

https://jira.carbonblack.local/browse/CBAPI-2285

## Pull Request Description

Implement the Sensor Lifecycle Management APIs and underlying support.

- [ ] Yes
- [x] No

## How Has This Been Tested?

Full unit tests have been added for all new code.  Coverage of new code is 100%.


src/cbc_sdk/workload/sensor_lifecycle.py
```diff
@@ -0,0 +1,210 @@
...
+from cbc_sdk.base import UnrefreshableModel
+
+import logging
+import json
+
+log = logging.getLogger(__name__)
+
+
+class SensorType(UnrefreshableModel):
+    """Represents a sensor type in a request."""
```
avanbrunt-cb, line 26: Not quite clear what is the request. How about this? `Represents a sensor type in a sensor installation request`

src/cbc_sdk/workload/sensor_lifecycle.py
```diff
@@ -0,0 +1,210 @@
...
+from cbc_sdk.errors import ApiError
+from cbc_sdk.base import UnrefreshableModel
+
+import logging
+import json
+
+log = logging.getLogger(__name__)
+
+
+class SensorType(UnrefreshableModel):
```
avanbrunt-cb, line 25: `SensorType` is confusing when you write `SensorType.create()` as you are not creating a new sensor type but instead a `SensorKit`.

src/cbc_sdk/workload/sensor_lifecycle.py
```diff
@@ -0,0 +1,210 @@
...
+
+        Returns:
+            SensorType: The sensor type information.
+        """
+        if self._sensor_type is None:
+            self._sensor_type = SensorType(self._cb, self._info['sensor_type'])
+        return self._sensor_type
+
+
+class SensorRequest:
```
avanbrunt-cb, line 115: This implementation is confusing where the two aspects are blended together and the SensorType model is not a model in the API it's a set of sub properties within the sensor model. My thoughts are that you should have a SensorKit model which combines the SensorType and SensorInfo along with a SensorKitQuery class similar to how the rest of the cbc sdk works.

`get_config_template` can be a classmethod on the SensorKit model

`add_sensor_type`, `expire_time`, and `config_params` would be functions on the SensorKitQuery then `get_sensor_info` would be replace by a `submit` function instead

`request_sensor_install` should be into the compute_resource model that Hristo created and then a class method could be used to support it in bulk. That way you could support a use case for the cbc sdk where a developer can run a function on the compute resource model `compute_resource.install_sensor(sensor_type, config_file=None)` or do it in bulk `ComputeResource.bulk_install(compute_resources, sensor_types, config_file=None)` Since the config file is optional in the request it can be optional in the function.


The above implementation would match closer to the rest of the cbc sdk and would integrate with the other models in the workloads package

src/cbc_sdk/connection.py
```diff
@@ -521,6 +521,24 @@ def post_object(self, uri, body, **kwargs):
...
+            uri (str): The URI to send the POST request to.
+            **kwargs (dict): Arguments to pass to the API. Except for "headers," these will all be added as parameters
+                             to the form data sent.
+
+        Returns:
+            object: The return data from the POST request.
+        """
+        headers = kwargs.pop("headers", {})
+        headers["Content-Type"] = "multipart/form-data"
+        print(kwargs)
```
emitreva-cb, line 538: Is this print added for a test?

src/cbc_sdk/workload/sensor_lifecycle.py
abowersox-cb: To do anything about this, I need two things.  First, I need Hristo's compute_resource model, which isn't in the code yet, but will be soon.  Second, I need you to explain more about this organization of functionality ("SensorKit"? Where did _that_ name come from?)  Until then, I'm at an impasse.

src/cbc_sdk/connection.py
abowersox-cb: I believe so.  That can be removed.

emitreva-cb (approved):

abowersox-cb:
@avanbrunt-cb, I'm going to want you to let me know if you want me to add any more sensor installing functions to ComputeResource.  Since they all ultimately go through the _do_sensor_install_request helper function, I can add other combinations of parameters if you want some extra syntactic sugar.  (Not too much though! Too much syntactic sugar causes cancer of the semicolon. :D )

src/cbc_sdk/workload/sensor_lifecycle.py
abowersox-cb: Please look at the new organization and see if it matches up with what you were saying.

src/cbc_sdk/workload/vm_workloads_search.py
```diff
@@ -82,6 +83,40 @@ def _build_api_request_uri(self, http_method="GET"):
         """
         return self.urlobject_single.format(self._cb.credentials.org_key, self._model_unique_id)
 
+    def install_sensor(self, sensor_kit_types, config_file=None):
```
avanbrunt-cb, line 86: Instead of a full sensor_kit_type can we just require the sensor version since the other properties are present in the compute resource to make the sensor_kit_type

src/cbc_sdk/workload/sensor_lifecycle.py
```diff
@@ -0,0 +1,253 @@
...
+    def __init__(self, cb, initial_data=None):
+        """
+        Initialize the SensorKit object.
+
+        Args:
+            cb (BaseAPI): Reference to API object used to communicate with the server.
+            initial_data (dict): Initial data used to populate the sensor kit data.
+        """
+        super(SensorKit, self).__init__(cb, None, initial_data)
+        self._full_init = True
```
avanbrunt-cb, line 43: This should not be considered _full_init unless the sensor kit url has been fetched i.e. `sensor_url` otherwise it's only a partial init

avanbrunt-cb:
> @avanbrunt-cb, I'm going to want you to let me know if you want me to add any more sensor installing functions to ComputeResource. Since they all ultimately go through the _do_sensor_install_request helper function, I can add other combinations of parameters if you want some extra syntactic sugar. (Not too much though! Too much syntactic sugar causes cancer of the semicolon. :D )

It might be worth supporting an array of objects with a vcenter_id and compute_resource_id for the bulk install incase they have the ids from somewhere else and don't have a compute_resource model

src/cbc_sdk/workload/__init__.py
```diff
@@ -1,3 +1,4 @@
 from cbc_sdk.workload.vm_workloads_search import ComputeResource
 from cbc_sdk.workload.vulnerability_assessment import (VulnerabilitySummary,
     OrganizationalVulnerability, Vulnerability, DeviceVulnerability)
+from cbc_sdk.workload.sensor_lifecycle import SensorKit, SensorKitQuery
```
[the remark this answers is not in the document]
abowersox-cb: Oops. Rectifying.

src/cbc_sdk/workload/vm_workloads_search.py
abowersox-cb: We discussed this but did not get a full resolution.

The logical place to look for this is in the `os_type` and `os_architecture` fields of the ComputeResource.  Which would be all well and good, except for the following:
* These fields are defined as _arrays._ How can or should I "smash them down" to the single value that SensorKit requires?
* There's not a 1:1 correspondence between the potential values of `os_type` and the allowed values for either `device_type` or `type` in SensorKit.  In particular, `os_type` appears to be _missing_ the "MAC" value as a valid value, which would seem to me to be a very serious omission.

Any advice on how to resolve these concerns would be appreciated.

src/cbc_sdk/workload/sensor_lifecycle.py
abowersox-cb: I'm putting in a workaround for this.  Let me know what you think when you see it.

src/cbc_sdk/workload/vm_workloads_search.py
avanbrunt-cb: I think that is a mistake in the schema portion of the documentation for compute resources as that appears to be defining the properties from the search criteria which supports more than one type.

See the response of the search or get
```{
    "id": "14101309",
    "appliance_uuid": "e66fdd46-2268-4f34-8a3f-226fa6d3edd0",
    "cluster_name": "cwp-bucket-1-cluster",
    "datacenter_name": "cwp-bucket-1-datacenter",
    "esx_host_name": "10.234.56.789",
    "esx_host_uuid": "a2311b42-3e53-8f21-97d7-66680007185f",
    "vcenter_name": "VMware vCenter Server 6.7.0 build-14368073",
    "vcenter_host_url": "10.234.56.789",
    "vcenter_uuid": "9a8a0be5-ae1e-49ce-b2aa-34bc7dc445e3",
    "name": "cwp-bucket-1-appliance",
    "host_name": "photon-machine",
    "created_at": "2020-10-29T16:37:17.610Z",
    "ip_address": "10.234.56.789",
    "eligibility": "UNSUPPORTED",
    "eligibility_code": [
        "Unsupported OS"
    ],
    "installation_status": "NOT_INSTALLED",
    "installation_status_code": null,
    "uuid": "500edcd7-e8ff-868a-6513-4e2fabcfcccd",
    "os_description": "VMware Photon OS (64-bit)",
    "os_type": "OTHER",
    "os_architecture": "32",
    "vmwaretools_version": "11296"
}
```

We should probably also verify the eligibility of the ComputeResource and throw an error if it is not supported.

MAC is likely missing because it is not a supported OS for vcenter so that should not be an issue

src/cbc_sdk/workload/sensor_lifecycle.py
avanbrunt-cb: That seems reasonable. Is there any chance that a model marked as pseudo would be updated or fetch the sensorkit and therefore making it a full initialized model?

abowersox-cb: I don't see how.  SensorKit is defined as an UnrefreshableModel, so you can't refresh it.  I don't think you can save it either, since there's no save mechanism defined.

avanbrunt-cb: Cool! Thanks for verifying

abowersox-cb:
> > @avanbrunt-cb, I'm going to want you to let me know if you want me to add any more sensor installing functions to ComputeResource. Since they all ultimately go through the _do_sensor_install_request helper function, I can add other combinations of parameters if you want some extra syntactic sugar. (Not too much though! Too much syntactic sugar causes cancer of the semicolon. :D )
> 
> It might be worth supporting an array of objects with a vcenter_id and compute_resource_id for the bulk install incase they have the ids from somewhere else and don't have a compute_resource model

How do you want that passed in? As a list of sicts with the two values `vcenter_id` and `compute_resource_id`? Would that make the most sense?

src/cbc_sdk/workload/vm_workloads_search.py
abowersox-cb: OK, I have dealt with this as well as the simple-pairs case mentioned in comments below.

avanbrunt-cb (approved):
This looks great! Thanks for making the changes I suggested

PR #129, opened 2021-02-16
Workloads Search example script

hkaragitliev-cb:
## Pull request checklist

https://jira.carbonblack.local/browse/CBAPI-2212

## Pull Request Description

Interactive console script which uses workloads search sdk 

- [ ] Yes
- [x] No

### DIFF
--- a/examples/workload/workloads_search_example.py
+++ b/examples/workload/workloads_search_example.py
@@ -1,0 +1,155 @@
+#!/usr/bin/env python3
+
+# *******************************************************
+# Copyright (c) VMware, Inc. 2021. All Rights Reserved.
+# SPDX-License-Identifier: MIT
+# *******************************************************
+# *
+# * DISCLAIMER. THIS PROGRAM IS PROVIDED TO YOU "AS IS" WITHOUT
+# * WARRANTIES OR CONDITIONS OF ANY KIND, WHETHER ORAL OR WRITTEN,
+# * EXPRESS OR IMPLIED. THE AUTHOR SPECIFICALLY DISCLAIMS ANY IMPLIED
+# * WARRANTIES OR CONDITIONS OF MERCHANTABILITY, SATISFACTORY QUALITY,
+# * NON-INFRINGEMENT AND FITNESS FOR A PARTICULAR PURPOSE.
+
+"""Example script showing VM Workloads Search operations.
+
+Docs: https://developer.carbonblack.com/reference/carbon-black-cloud/workload-protection/latest/vm-workload-search/
+"""
+
+# Standard library imports
+import sys
+
+# Internal library imports
+from cbc_sdk.helpers import build_cli_parser
+from cbc_sdk.helpers import get_cb_cloud_object
+from cbc_sdk.workload.vm_workloads_search import ComputeResource
+
+
+def generate_menu():
+    """Define console menu"""
+    menu_content = {'0': {'name': 'M A I N - M E N U\n',
+                          'function_call': main},
+                    '1': {'name': 'Fetch Compute Resource by ID',
+                          'function_call': fetch_compute_resource_by_id},
+                    '2': {'name': 'Search and Facet Compute Resources',
+                          'function_call': search_and_facet_compute_resources},
+                    '9': {'name': 'Quit',
+                          'function_call': quit_script}
+                    }
+
+    return menu_content
+
+
+def user_input(menu):
+    """Prompt user for menu choice"""
+    print('-' * 21)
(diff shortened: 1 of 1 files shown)


hkaragitliev-cb:
I really don't like using eval(), but i couldn't figure any other way of using multiple set_ values. Passing empty values does not work.

examples/workload/workloads_search_example.py
```diff
@@ -0,0 +1,155 @@
...
+        if args.ip_address:
+            query += '.set_ip_address(["' + '","'.join(map(str, args.ip_address)) + '"])'
+        if args.installation_status:
+            query += '.set_installation_status(["' + '","'.join(map(str, args.installation_status)) + '"])'
+        if args.uuid:
+            query += '.set_uuid(["' + '","'.join(map(str, args.uuid)) + '"])'
+        if args.os_architecture:
+            query += '.set_os_architecture(["' + '","'.join(map(str, args.os_architecture)) + '"])'
+        if args.eligibility:
+            query += '.set_eligibility(["' + '","'.join(map(str, args.eligibility)) + '"])'
```
avanbrunt-cb, line 113: Yeah this is going to have to change. The following code should work as each set function will return the updated query class
```
query = cbc.select(ComputeResource)
    use_filter = input('\nWould you like to use criteria filters?: Y/n\n')
    if use_filter.lower() in ['y', 'yes']:
        print('Each criteria accepts one or more space separated values.',
              'Exmaple: --name ABCD --appliance_uuid 1234 5678')
        args = parser.parse_args(input('Enter criteria filters: ').split())
        if args.name:
            query = query.set_name(["' + '","'.join(map(str, args.name)) + '"])
        if args.os_type:
            query = query.set_os_type(["' + '","'.join(map(str, args.os_type)) + '"])
        if args.appliance_uuid:
            query = query.set_appliance_uuid(["' + '","'.join(map(str, args.appliance_uuid)) + '"])
        if args.cluster_name:
            query = query.set_cluster_name(["' + '","'.join(map(str, args.cluster_name)) + '"])
        if args.ip_address:
            query = query.set_ip_address(["' + '","'.join(map(str, args.ip_address)) + '"])
        if args.installation_status:
            query = query.set_installation_status(["' + '","'.join(map(str, args.installation_status)) + '"])
        if args.uuid:
            query = query.set_uuid(["' + '","'.join(map(str, args.uuid)) + '"])
        if args.os_architecture:
            query = query.set_os_architecture(["' + '","'.join(map(str, args.os_architecture)) + '"])
        if args.eligibility:
            query = query.set_eligibility(["' + '","'.join(map(str, args.eligibility)) + '"])
```

hkaragitliev-cb: for some reason, it did not occur to me that I can use the filters after the query is made...

emitreva-cb (approved):

examples/workload/workloads_search_example.py
```diff
@@ -0,0 +1,156 @@
...
+        if args.installation_status:
+            query.set_installation_status(args.installation_status)
+        if args.uuid:
+            query.set_uuid(args.uuid)
+        if args.os_architecture:
+            query.set_os_architecture(args.os_architecture)
+        if args.eligibility:
+            query.set_eligibility(args.eligibility)
+
+    print(*query)
```
avanbrunt-cb, line 116: Does this execute the query or just print the query in string format? If it only prints the query then we should probably do loop here to print the compute resources that were found from the search

hkaragitliev-cb: It executes it. And the best part is that you can stack any number of filters.

avanbrunt-cb: That's cool! Though I don't think you are correct with `the execution is done with the query.set_eligibility(args.eligibility)` as this area of the code does not perform the execution of the query https://github.com/carbonblack/carbon-black-cloud-sdk-python/blob/develop/src/cbc_sdk/workload/vm_workloads_search.py#L127

What I believe happens is the *query is likely treated as a iter-able function so the print statement loops through potentially.

Can you verify that this will work for more than one compute resource?

hkaragitliev-cb: https://user-images.githubusercontent.com/74309356/108233444-750a6200-714c-11eb-8219-3310b56188f6.mov

avanbrunt-cb: Awesome! I like the use of the video for proof

avanbrunt-cb (approved):
Looks awesome and works really well!

PR #137, opened 2021-03-10
CBAPI-2022: Process Summary Search

emitreva-cb:
## Pull request checklist

https://jira.carbonblack.local/browse/CBAPI-2022

## Pull Request Description
- increased the code coverage of platform/process

- [ ] Yes
- [x] No


src/tests/unit/platform/test_platform_process.py
```diff
@@ -86,10 +96,119 @@ def test_process_select(cbcsdk_mock):
...
+    cbcsdk_mock.mock_request("GET", ("/api/investigate/v2/orgs/test/processes/"
+                                     "summary_jobs/2c292717-80ed-4f0d-845f-779e09470920"),
+                             GET_PROCESS_SUMMARY_RESP_STILL_QUERYING)
     api = cbcsdk_mock.api
     guid = 'WNEXFKQ7-0002b226-000015bd-00000000-1d6225bbba74c00'
     summary = api.select(Process.Summary).where(f"process_guid:{guid}")
+    assert summary._still_querying() is True
+    summary._search()
+    summary._timed_out = True
+    summary._search()
```
avanbrunt-cb, line 176: This needs an assert or some kind of verification that having set the _timed_out the following search function call did something different or completed

emitreva-cb: Actually I tried to cover one part of the _search() function, but this did not do the trick for some reason, so removing the last two lines - there are not doing anything.

src/cbc_sdk/platform/processes.py
```diff
@@ -728,27 +815,6 @@ def _still_querying(self):
 
         return False
 
-    def _count(self):
```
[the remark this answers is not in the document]
emitreva-cb: Good point, adding the _count with exception that the result is not iterable as the summary returns just a single record.

src/tests/unit/platform/test_platform_process.py
```diff
@@ -78,6 +87,7 @@ def test_process_select(cbcsdk_mock):
     api = cbcsdk_mock.api
     guid = 'WNEXFKQ7-0002b226-000015bd-00000000-1d6225bbba74c00'
     process = api.select(Process, guid)
+    process.summary.__str__()
```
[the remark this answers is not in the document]
emitreva-cb: OK, adding for both summary and tree.

src/cbc_sdk/platform/processes.py
```diff
@@ -683,14 +760,19 @@ def _get_query_parameters(self):
...
+            args["process_guid"] = q
+
         if 'parent_guid:' in query:
             # extract parent_guid from where() clause
             parent_guid = query.split('parent_guid:', 1)[1].split(' ', 1)[0]
             args["parent_guid"] = parent_guid
         return args
 
+    def _count(self):
+        raise ApiError('The result is not iterable')
```
avanbrunt-cb, line 774: As much as this is the case the SummaryQuery class inherits IterableQueryMixin so this message is confusing. We should consider removing the IterableQueryMixin inheritance as well.

src/tests/unit/platform/test_platform_process.py
avanbrunt-cb: @emitreva-cb There is still no validation after `summary._search()`. Removing the two lines is a start however there needs to be some verification of what that line is doing. Either a comment saying this will raise an error if there is a mistake/bug or verify that the search completed successfully

avanbrunt-cb (approved):

hkaragitliev-cb (approved):

Issue #144, opened 2021-03-25
[BUG] - Duplicates in "list_devices.py" example


MarkTheTiger:
I am seeing this behaviour on "list_devices.py" example (Platform Section)

OS: WIndows Server 2016
Carbon Black Cloud Products: Audit & Remediation
Python Version: 3.9.1

Describe the bug :
When results are printed to screen (showing the list of the devices) it randomly duplicates an entry, but keeping the total number of devices correct, so some devices are not shown and instead other devices are duplicated.
I've included the script in a scheduled task to refresh every 10 minutes the devices list, and I noticed that duplicated devices are never the same.

avanbrunt-cb:
@MarkTheTiger Can you provide the CLI you are using which causes the issue you have mentioned? It's possible that if you are sorting on a property that is updated when we page through the results it could be adjusting the list which would cause a Device to be outputted twice

MarkTheTiger:
Hi @avanbrunt-cb,
I took the example "list_devices.py" and runned it directly from Atom recalling Python (using atom-python-run, without a CLI).
After that, I also launched the script via Powershell 5.1 .

avanbrunt-cb:
Did you provide any arguments to the script?

MarkTheTiger:
No, just ".\list_devices.py" in powershell, in atom runned pressing F5.

avanbrunt-cb:
@MarkTheTiger I've found that the devices request applies a sort on `last_contact_time` by default so because of our current pagination if you have a few thousand devices it takes a little bit of extra time to process all of the devices. I have a workaround for now but will look into adding a `set_rows` function to allow this to be more easily configured.

The workaround is to change the line 56 to the following
```
devices = list(query._perform_query(max_rows=10000))
```

This will set the query to retrieve the max number of devices available in one request instead of paginating.

avanbrunt-cb:
Fixed in release v1.2.1. You can now add `query.set_max_rows(10000)` to also fetch all devices in a single query instead of using the default pagination.

PR #148, opened 2021-03-30
Add set_max_rows for device query

avanbrunt-cb:
## Pull request checklist

#144 

## Pull Request Description

Add set_max_rows to device query

- [ ] Yes
- [x] No

## How Has This Been Tested?

Unit test added

### DIFF
--- a/src/cbc_sdk/platform/devices.py
+++ b/src/cbc_sdk/platform/devices.py
@@ -267,6 +267,7 @@ class DeviceSearchQuery(BaseQuery, QueryBuilderSupportMixin, CriteriaBuilderSupp
         self._time_filter = {}
         self._exclusions = {}
         self._sortcriteria = {}
+        self.max_rows = -1
 
     def _update_exclusions(self, key, newlist):
         """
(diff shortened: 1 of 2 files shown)


src/cbc_sdk/platform/devices.py
```diff
@@ -485,6 +486,24 @@ def set_deployment_type(self, deployment_type):
...
+            max_rows (integer): Max number of devices
+
+        Returns:
+            DeviceSearchQuery: This instance.
+
+        Raises:
+            ApiError: If rows is negative or greater than 10000
+        """
+        if max_rows < 0 or max_rows > 10000:
+            raise ApiError("Max rows must be between 0 and 10000")
```
mikethecanuck-cb, line 503: What's the benefit of setting `max_rows = 0`? An API response with no returned rows - does that do anything practically useful, or just theoretically interesting?

mikethecanuck-cb (approved):
Looks good, assuming that `rows = 0` is something that we need to or should support

src/cbc_sdk/platform/devices.py
avanbrunt-cb: It's a supported value and could be used to validate a search or check num_found without fetching any results. May not be practical but it's not something I want to prevent

emitreva-cb (approved):

PR #149, opened 2021-03-30
CBAPI-2351: Access Profiles - SDK Code Changes

abowersox-cb:
## Pull request checklist

https://jira.carbonblack.local/browse/CBAPI-2351

## Pull Request Description

Access profiles and grants APIs for CBC SDK.

- [ ] Yes
- [x] No

## How Has This Been Tested?

Unit tests have been added for all new code with ~~93%~~ ~~94%~~ _95%_ coverage of the new code.


emitreva-cb (approved):

emitreva-cb:
Just wondering about something - if cidr and expiration are not implemented (in the APIs), should those be added to the SDK - or it is easier once they are implemented, it will not be needed to update this implementation afterwards to include them?

src/cbc_sdk/platform/admin_grants.py
```diff
@@ -0,0 +1,471 @@
...
+                Profile: The new Profile object.
+            """
+            data = {'orgs': self._orgs, 'roles': self._roles, 'conditions': self._conditions,
+                    'can_manage': self._can_manage}
+            profile = Grant.Profile(self._grant._cb, self._grant, None, data)
+            self._grant._profiles.append(profile)
+            self._grant.touch()
+            return profile
+
+    def __init__(self, cb, model_unique_id, initial_data=None):
```
avanbrunt-cb, line 261: From an object structure do you mind putting the `__init__` function before the sub class Profile

src/cbc_sdk/platform/admin_grants.py
```diff
@@ -0,0 +1,471 @@
...
+    def profiles_(self):
+        """
+        Return the profiles associated with this grant.
+
+        Returns:
+            list: The profiles associated with this grant, each represented as a Profile object.
+        """
+        return self._profiles
+
+    def new_profile(self):
```
avanbrunt-cb, line 368: Can we call the function `create_profile` and add an optional object where a user can specify an object for the profile instead of using the ProfileBuilder. This will let a user to have both as an option in case they already have a profile defined

src/cbc_sdk/platform/admin_grants.py
```diff
@@ -0,0 +1,550 @@
...
+        return update_return
+
+    def _delete_object(self):
+        """Deletes the grant from the principal."""
+        url = self.urlobject_single.format(self._cb.credentials.org_key, self._model_unique_id)
+        ret = self._cb.delete_object(url)
+        self._refresh_if_needed(ret)
+
+    @classmethod
+    def new_grant(cls, cb, principal):
```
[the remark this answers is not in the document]
abowersox-cb: Changing this.

src/cbc_sdk/platform/admin_grants.py
```diff
@@ -0,0 +1,550 @@
...
+            Sets the management flag for the new profile.
+
+            Args:
+                flag (bool): New value of the management flag.
+
+            Returns:
+                ProfileBuilder: This object.
+            """
+            self._can_manage = flag
+            return self
```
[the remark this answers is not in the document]
abowersox-cb: OK, this will mainly affect the `ProfileBuilder`.

src/cbc_sdk/platform/admin_grants.py
```diff
@@ -0,0 +1,550 @@
...
+
+        Returns:
+            Grant: The new grant object.
+        """
+        grant = Grant(cb, principal, {'principal': principal})
+        grant._full_init = True
+        return grant
+
+    @property
+    def profiles_(self):
```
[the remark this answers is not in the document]
abowersox-cb: Because "profiles" is the name used by the raw data.  This matches the convention we use elsewhere.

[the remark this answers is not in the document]
avanbrunt-cb: Thanks for the explanation this makes sense

src/cbc_sdk/platform/admin_grants.py
```diff
@@ -0,0 +1,550 @@
...
+            Args:
+                flag (bool): New value of the management flag.
+
+            Returns:
+                ProfileBuilder: This object.
+            """
+            self._can_manage = flag
+            return self
+
+        def build(self):
```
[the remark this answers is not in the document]
abowersox-cb: This is a different pattern.  The `new_profile()` method on the `Grant` returns a `ProfileBuilder`, which is used to set up all the data that the new profile will contain.  The `build()` method then takes that data and uses it to build an actual `Profile`.  I don't know if we use this pattern elsewhere in the SDK, but it's one that I'm familiar with in other contexts.

src/cbc_sdk/platform/admin_grants.py
```diff
@@ -0,0 +1,550 @@
...
+
+            Returns:
+                bool: True if refresh was successful, False if not.
+            """
+            url = self.urlobject_single.format(self._cb.credentials.org_key, self._grant._model_unique_id,
+                                               self._model_unique_id)
+            resp = self._cb.get_object(url)
+            self._info = resp
+            self._last_refresh_time = time.time()
+            return True
```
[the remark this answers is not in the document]
abowersox-cb: No, here we use GET `/grants/{principal}/profiles/{profile_uuid}`.  There are three parameters in the string `urlobject_single`.  This is why we have the backreference to the `Grant` from inside the `Profile`.

src/cbc_sdk/platform/admin_grants.py
```diff
@@ -0,0 +1,471 @@
...
+            Adds the specified organization to the list of organizations for which the new profile is allowed.
+
+            Args:
+                org (str): URN of the organization to be added.
+
+            Returns:
+                ProfileBuilder: This object.
+            """
+            self._orgs['allow'].append(org)
+            return self
```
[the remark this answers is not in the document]
abowersox-cb: We already have `set_orgs()`, that sets the entire "orgs" data structure. Given that that structure only now contains one list, though, that method could be deleted and the name repurposed.

src/cbc_sdk/platform/admin_grants.py
abowersox-cb: I think that actually _muddles_ the object structure to have `__init__` away from the rest of the `Profile` code.  But we'll do it your way.

src/cbc_sdk/platform/admin_grants.py
abowersox-cb: I can change the name, but why would they already have a profile defined?

src/cbc_sdk/platform/admin_grants.py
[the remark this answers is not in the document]
avanbrunt-cb: Given this is a new builder class I am open to using build but we should be cognizant of existing methods in the CBC SDK and reuse existing ones over creating new mechanisms. This seems like a good one to create new though given create is not something we have done much off.

src/cbc_sdk/platform/admin_grants.py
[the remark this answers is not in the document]
avanbrunt-cb: That VERB is supported on that endpoint.

src/cbc_sdk/platform/admin_grants.py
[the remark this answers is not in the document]
avanbrunt-cb: Yep! That's what I was thinking as well

src/cbc_sdk/platform/admin_grants.py
avanbrunt-cb: I know this is personal preference but I prefer being able to see how the variables are initialized next to the definition

src/cbc_sdk/platform/admin_grants.py
avanbrunt-cb: In an automation scenario they will already have their profile defined and just need to apply it to each grant they create for each new user. This will let them do that in one go instead of needing to call multiple functions in the ProfileBuilder

src/cbc_sdk/platform/admin_grants.py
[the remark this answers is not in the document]
abowersox-cb: This has been done.

src/cbc_sdk/platform/admin_grants.py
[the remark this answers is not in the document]
abowersox-cb: You mean it's _not_ supported on that endpoint. As you told me via Slack. The `_refresh()` has been properly gutted.

src/cbc_sdk/platform/admin_grants.py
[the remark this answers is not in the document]
abowersox-cb: Besides which, since you told me to change `new_profile()` to `create_profile()`, that means we would have, effectively, a control flow like `grant.create_profile().{builder operations}.create()`. Using "create" twice would be confusing.

src/cbc_sdk/platform/admin_grants.py
abowersox-cb: Already done.

src/cbc_sdk/platform/admin_grants.py
abowersox-cb: This has been added.

src/cbc_sdk/platform/admin_grants.py
```diff
@@ -0,0 +1,446 @@
...
+
+        Args:
+            cb (CBCloudAPI): A reference to the CBCloudAPI object.
+            principal (str): The principal URN for the new grant object.
+
+        Returns:
+            Grant: The new grant object.
+        """
+        grant = Grant(cb, principal, {'principal': principal})
+        grant._full_init = True
```
avanbrunt-cb, line 320: This is not correct and misleads the current object. The Grant is not `_full_init`

src/cbc_sdk/platform/admin_grants.py
```diff
@@ -0,0 +1,446 @@
...
+        return update_return
+
+    def _delete_object(self):
+        """Deletes the grant from the principal."""
+        url = self.urlobject_single.format(self._cb.credentials.org_key, self._model_unique_id)
+        ret = self._cb.delete_object(url)
+        self._refresh_if_needed(ret)
+
+    @classmethod
+    def create(cls, cb, principal):
```
avanbrunt-cb, line 308: There should be an optional object here to provide the full grant object similar to how profiles is being done.

src/tests/unit/platform/test_admin_grants.py
```diff
@@ -0,0 +1,209 @@
...
+        ret['profile_uuid'] = body['profile_uuid']
+        return ret
+
+    cbcsdk_mock.mock_request('GET', '/access/v2/orgs/test/grants/psc:user:12345678:ABCDEFGH', GET_GRANT_RESP)
+    cbcsdk_mock.mock_request('POST', '/access/v2/orgs/test/grants/psc:user:ABC12345:DEF67890/profiles',
+                             respond_to_profile_grant)
+    api = cbcsdk_mock.api
+    grant = Grant(api, 'psc:user:12345678:ABCDEFGH')
+    template = {
+        "profile_uuid": "to-be-deleted",
```
avanbrunt-cb, line 123: The profile_uuid should not be specified in the template as that is created by the backend

src/tests/unit/platform/test_admin_grants.py
```diff
@@ -0,0 +1,209 @@
...
+    """Test the deletion of a grant."""
+    cbcsdk_mock.mock_request('GET', '/access/v2/orgs/test/grants/psc:user:12345678:ABCDEFGH', GET_GRANT_RESP)
+    cbcsdk_mock.mock_request('DELETE', '/access/v2/orgs/test/grants/psc:user:12345678:ABCDEFGH', DELETE_GRANT_RESP)
+    api = cbcsdk_mock.api
+    grant = Grant(api, 'psc:user:12345678:ABCDEFGH')
+    grant.delete()
+    assert grant.revoked is True
+
+
+def test_query_grants(cbcsdk_mock):
```
avanbrunt-cb, line 178: I don't see a test for getting a Grant by ID using the select like so `api.select(Grant, 'principal_urn')`

avanbrunt-cb (commented):
Please rename the `admin_grants.py` file to `grants.py` as we don't want to reference `admin` as this applies to `users`

src/tests/unit/platform/test_admin_grants.py
abowersox-cb: That's because there isn't one. Does there need to be one, given that the code it would run has already been adequately covered by other tests?

src/tests/unit/platform/test_admin_grants.py
abowersox-cb: No, it shouldn't be, and, if you look at the code for `create_profile()`, the profile_uuid gets explicitly stripped out if it exists.  We have it in the template for this test so that the stripping-out can be tested.

avanbrunt-cb: Then please add a comment indicating that this is being tested this way

src/tests/unit/platform/test_admin_grants.py
avanbrunt-cb: You could adjust `test_get_and_set_grant` to use this method to get this usecase

src/cbc_sdk/platform/admin_grants.py
abowersox-cb: If I take it out, two of my tests fail with the following error:
```
Failed: GET called for /access/v2/orgs/test/grants/psc:user:ABC12345:DEF67890 when it shouldn't be
```
There's a sneak path in the `__setattr__` code in the `MutableBaseModel` that is causing `refresh()` to be called when it shouldn't be.

avanbrunt-cb: Then we should implement an alternative approach to solve this issue. Let's not mess with existing practices as this may cause more issues in the future

src/tests/unit/platform/test_admin_grants.py
abowersox-cb: Adding this.

src/tests/unit/platform/test_admin_grants.py
abowersox-cb: OK

src/cbc_sdk/platform/admin_grants.py
abowersox-cb: What do you suggest? I was racking my brains for _hours_ to find an alternate way of solving this, to no avail.

src/cbc_sdk/platform/admin_grants.py
abowersox-cb: Adding.

src/cbc_sdk/platform/admin_grants.py
avanbrunt-cb: I see three options with varying types of UX:
* Require the entire grant like we have done for reputation https://github.com/carbonblack/carbon-black-cloud-sdk-python/blob/develop/src/cbc_sdk/platform/reputation.py#L93
* Create a Builder like we have for Profiles
* Lastly, if the following code is the experience we desire then we need to add a new property similar to _full_init that will allow for additional new behavior in `def _set(self, attrname, new_value)`. A possible name could be `_created`
```
grant = Grant.create(api, 'psc:user:ABC12345:DEF67890')
grant.roles = ["psc:role::SECOPS_ROLE_MANAGER"]
grant.org_ref = 'psc:org:test'
grant.principal_name = 'Doug Jones'
```

src/cbc_sdk/platform/grants.py
```diff
@@ -141,25 +148,25 @@ def set_orgs(self, orgs_list):
...
+            self._orgs['allow'] = [f"psc:org:{org}" for org in orgs_list]
             return self
 
         def add_org(self, org):
             """
             Adds the specified organization to the list of organizations for which the new profile is allowed.
 
             Args:
-                org (str): URN of the organization to be added.
+                org (str): Name of the organization to be added.
```
avanbrunt-cb, line 164: This is the `org_key` not the `name`

src/cbc_sdk/platform/grants.py
```diff
@@ -141,25 +148,25 @@ def set_orgs(self, orgs_list):
             Set the list of organizations to which the new profile is allowed access.
 
             Args:
-                orgs_list (list): List of organization URNs.
+                orgs_list (list): List of organization names.
```
avanbrunt-cb, line 151: `org_key` not `name`

src/cbc_sdk/platform/grants.py
```diff
@@ -280,22 +396,30 @@ def _update_object(self):
...
+            save_info['profiles'] = [copy.deepcopy(prof._info)
+                                     for prof in self._profiles if 'profile_uuid' in prof._info]
             url = self.urlobject_single.format(self._cb.credentials.org_key, save_info['principal'])
             ret = self._cb.put_object(url, save_info)
         else:
             log.debug("Creating a new Grant object for principal {0:s}".format(save_info['principal']))
+            save_info['profiles'] = []
+            for profile in self._profiles:
+                # presumed to be all new profiles, so create UUIDs for them
+                profile._info['profile_uuid'] = str(uuid.uuid4())
```
avanbrunt-cb, line 412: Have you confirmed if this is necessary to create your own uuid? On the creation of a grant will it accept profiles without uuids?

src/cbc_sdk/platform/grants.py
```diff
@@ -305,26 +429,34 @@ def _delete_object(self):
...
-            t['principal'] = principal
-        else:
-            t = {'principal': principal}
-        grant = Grant(cb, principal, t)
-        grant._full_init = True  # XXX Alex says this is wrong, but I'm keeping it until we have a better solution
-        return grant
+            grant = Grant(cb, t['principal'], t)
+            grant._update_object()
+            return grant
+        if not all([key in kwargs for key in ['orgid', 'userid']]):
```
avanbrunt-cb, line 457: `orgid` should be `org_key`

avanbrunt-cb (commented):
Couple small changes. Has this been tested manually against a Carbon Black Cloud instance?

src/tests/unit/platform/test_admin_grants.py
```diff
@@ -0,0 +1,280 @@
+"""Tests for the Grant and Profile objects of the CBC SDK"""
```
avanbrunt-cb, line 1: This file should be renamed `test_grants.py` to match the other file rename

src/cbc_sdk/platform/grants.py
abowersox-cb: Fixing this.

src/cbc_sdk/platform/grants.py
abowersox-cb: Also this.

src/tests/unit/platform/test_admin_grants.py
abowersox-cb: Doing this and leaving my personal feelings out of it.

avanbrunt-cb (approved):
Thanks for all the effort for working through the suggested changes

PR #157, opened 2021-04-16
CBAPI-2360: User Management SDK Code Changes

abowersox-cb:
## Pull request checklist

https://jira.carbonblack.local/broswe/CBAPI-2360

## Pull Request Description

Implement User Managemett: the User model object and associated query class.

- [ ] Yes
- [x] No

## How Has This Been Tested?

Unit tests have been written with 97% coverage for all new code.  The code has also been run against the dev01 server with a simple script designed to exercise the GET and PUT of a user.  (Script to be included in a separate comment.)

### DIFF
--- a/src/cbc_sdk/platform/users.py
+++ b/src/cbc_sdk/platform/users.py
@@ -204,6 +204,7 @@ class User(MutableBaseModel):
         if len(rawdata) == 0:
             return False
         self._info = rawdata[0]
+        self._full_init = True
         self._last_refresh_time = time.time()
         return True
 
--- a/src/cbc_sdk/platform/users.py
+++ b/src/cbc_sdk/platform/users.py
@@ -13,7 +13,7 @@
 
 """Model and Query Classes for Users"""
 from cbc_sdk.base import MutableBaseModel, BaseQuery, IterableQueryMixin, AsyncQueryMixin
-from cbc_sdk.errors import ApiError, ServerError
+from cbc_sdk.errors import ApiError, ServerError, ObjectNotFoundError
 import time
 import copy
 
@@ -31,7 +31,7 @@ def normalize_org(org):
 class User(MutableBaseModel):
     """Represents a user in the Carbon Black Cloud."""
     urlobject = "/appservices/v6/orgs/{0}/users"
-    urlobject_single = "/appservices/v6/orgs/{0}/users{1}"
+    urlobject_single = "/appservices/v6/orgs/{0}/users/{1}"
     primary_key = "login_id"
     swagger_meta_file = "platform/models/user.yaml"
 
@@ -46,7 +46,8 @@ class User(MutableBaseModel):
         """
         super(User, self).__init__(cb, model_unique_id, initial_data)
         if model_unique_id is not None and initial_data is None:
-            self._refresh()
+            if not self._refresh():
+                raise ObjectNotFoundError(f"user with login ID {model_unique_id} not found")
 
     class UserBuilder:
         """Auxiliary object used to construct a new User."""
@@ -58,8 +59,7 @@ class User(MutableBaseModel):
                 cb (BaseAPI): Reference to API object used to communicate with the server.
             """
             self._cb = cb
-            self._criteria = {'org_id': 0, 'role': 'DEPRECATED', 'auth_method': 'PASSWORD',
-                              'add_without_invite': False}
+            self._criteria = {'org_id': 0, 'role': 'DEPRECATED', 'auth_method': 'PASSWORD'}
 
         def set_email(self, email):
             """
@@ -71,7 +71,7 @@ class User(MutableBaseModel):
             Returns:
                 UserBuilder: This object.
             """
-            self._criteria['email'] = email
+            self._criteria['email_id'] = email
             return self
 
         def set_role(self, role):
@@ -126,19 +126,6 @@ class User(MutableBaseModel):
             self._criteria['auth_method'] = method
             return self
 
-        def set_send_invitation(self, invite):
-            """
-            Sets whether or not an invitation will be sent to the new user via E-mail.
-
-            Args:
-                invite (bool): True to have an invitation sent, False to not have one sent.
-
-            Returns:
-                UserBuilder: This object.
-            """
-            self._criteria['add_without_invite'] = False if invite else True
-            return self
-
         def add_grant_profile(self, orgs, roles):
             """
             Adds a grant profile for the new user.
@@ -150,11 +137,11 @@ class User(MutableBaseModel):
             Returns:
                 UserBuilder: This object.
             """
-            grant_prof = {'orgs': {'allow': [normalize_org(org) for org in orgs]}, 'roles': roles}
-            if 'grant_profiles' in self._criteria:
-                self._criteria['grant_profiles'].append(grant_prof)
+            new_profile = {'orgs': {'allow': [normalize_org(org) for org in orgs]}, 'roles': roles}
+            if 'profiles' in self._criteria:
+                self._criteria['profiles'].append(new_profile)
             else:
-                self._criteria['grant_profiles'] = [grant_prof]
+                self._criteria['profiles'] = [new_profile]
             return self
 
         def build(self):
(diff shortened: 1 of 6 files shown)


abowersox-cb:
User exercising script:
```
from cbc_sdk.platform import User
import time
import random
import argparse


parser = argparse.ArgumentParser(description='Rename a random user that has no name.')
parser.add_argument('first_name', help='First name to give the user.')
parser.add_argument('last_name', help='Last name to give the user.')
args = parser.parse_args()
print(f"I want to give the user a name {args.first_name} {args.last_name}")

api = CBCloudAPI(profile='dev01')
query = api.select(User)
start_time = time.time()
list_users = list(query)
print(f"Retrieved {len(list_users)} users from server; total time {time.time() - start_time}")

random.shuffle(list_users)
did_it = False
for user in list_users:
    if user.first_name == '' and user.last_name == '':
        print(f"Changing name of user with login ID {user.login_id} and email {user.email}")
        print(f"User dump: {user._info}")
        user.first_name = args.first_name
        user.last_name = args.last_name
        print(f"Stored user name {user.first_name} {user.last_name}.")
        user.save()
        print(f"That user's name is now {user.first_name} {user.last_name}.")
        did_it = True
        break

if not did_it:
    print("Oops, couldn't find any users without a name.")
```

abowersox-cb:
Note to future implementors: The `_refresh()` method of `NewBaseModel` sets the `_full_init` flag on the object to `True`, and there are other parts of that class and `MutableBaseModel` that rely on this.  So, if you are implementing a subclass and override `_refresh()`, your implementation _must_ set `_full_init` to `True` as well.

emitreva-cb (approved):

src/cbc_sdk/platform/users.py
```diff
@@ -0,0 +1,342 @@
...
+            my_templ['org_id'] = 0
+            my_templ['role'] = 'DEPRECATED'
+            if 'auth_method' not in my_templ:
+                my_templ['auth_method'] = 'PASSWORD'
+            return User._create_user(cb, my_templ)
+        return User.UserBuilder(cb)
+
+    def reset_google_authenticator_registration(self):
+        """Forces Google Authenticator registration to be reset for this user."""
+        url = self.urlobject_single.format(self._cb.credentials.org_key, self._model_unique_id) + "/googleauth"
```
hkaragitliev-cb, line 266: We received an updated swagger, there was a mistake in the original. The URL for the reset is `/google-auth`, instead of `/googleauth`

src/tests/unit/platform/test_users.py
```diff
@@ -0,0 +1,217 @@
...
+
+def test_get_nonexistent_user(cbcsdk_mock):
+    """Test retrieving a nonexistent user."""
+    cbcsdk_mock.mock_request('GET', '/appservices/v6/orgs/test/users', GET_USERS_RESP)
+    api = cbcsdk_mock.api
+    with pytest.raises(ObjectNotFoundError):
+        api.select(User, 999)
+
+
+def test_attempt_to_create_naively(cb):
```
avanbrunt-cb, line 122: Is this a behavior we want to prevent or should we support this by checking if an ID exists and if it doesn't then create a new User?

src/tests/unit/platform/test_users.py
```diff
@@ -0,0 +1,217 @@
...
+        return GET_USERS_AFTER_CREATE_RESP
+
+    cbcsdk_mock.mock_request('POST', '/appservices/v6/orgs/test/users', check_post)
+    cbcsdk_mock.mock_request('GET', '/appservices/v6/orgs/test/users', check_get)
+    api = cbcsdk_mock.api
+    builder = User.create(api).set_email('rios@la-sirena.net').set_role('psc:role:test:APP_SERVICE_ROLE')
+    builder.set_first_name('Cristobal').set_last_name('Rios')
+    builder.add_grant_profile(['psc:org:test2'], ['psc:role:test2:DUMMY'])
+    builder.add_grant_profile(['test3'], ['psc:role:test3:DUMMY'])
+    user, password = builder.build()
```
avanbrunt-cb, line 155: Have we confirmed this behavior as my understanding is that the creation of the user requires the user to receive the email and fill out their own password. I am not aware of a temporary password that is useable

src/tests/unit/platform/test_users.py
```diff
@@ -0,0 +1,217 @@
...
+        nonlocal post_made
+        assert post_made
+        return GET_USERS_AFTER_CREATE_RESP
+
+    cbcsdk_mock.mock_request('POST', '/appservices/v6/orgs/test/users', check_post)
+    cbcsdk_mock.mock_request('GET', '/appservices/v6/orgs/test/users', check_get)
+    api = cbcsdk_mock.api
+    template = {
+        "email_id": "rios@la-sirena.net",
+        "role_urn": "psc:role:test:APP_SERVICE_ROLE",
```
avanbrunt-cb, line 184: A grant can't have both a role_urn and a profile, it's one or the other

src/tests/unit/platform/test_users.py
kebringer-cb: Correct.

src/tests/unit/platform/test_users.py
abowersox-cb: This is _test code,_ intended to see that the data gets slotted-in correctly.  Does that matter? If it does, I'll have to write extra tests.

src/tests/unit/platform/test_users.py
abowersox-cb: Taking out return of the password, as per @kebringer-cb 's discussion.

src/tests/unit/platform/test_users.py
abowersox-cb: We don't support it for Grant; we require that they go through the Builder or use a template. This is by analogy.

src/cbc_sdk/platform/users.py
```diff
@@ -0,0 +1,342 @@
...
+        """Auxiliary object used to construct a new User."""
+        def __init__(self, cb):
+            """
+            Create the empty UserBuilder object.
+
+            Args:
+                cb (BaseAPI): Reference to API object used to communicate with the server.
+            """
+            self._cb = cb
+            self._criteria = {'org_id': 0, 'role': 'DEPRECATED', 'auth_method': 'PASSWORD'}
```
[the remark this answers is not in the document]
abowersox-cb: Renaming it to `_creation_data`. Will that be sufficient?

src/cbc_sdk/platform/users.py
abowersox-cb: Got that changed, thanks!

src/tests/unit/platform/test_users.py
avanbrunt-cb: Then let's change the test name to be `test_unsupported_create_by_update` as the language here is not appropriate in my opinion

src/tests/unit/platform/test_users.py
avanbrunt-cb: Test code must follow the same usage patterns otherwise this may cause confusion in the future or a developer may look at test code and make a wrong assumption

src/tests/unit/platform/test_users.py
abowersox-cb: As you wish.

src/cbc_sdk/platform/users.py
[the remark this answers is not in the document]
avanbrunt-cb: That's fine with me just not `_criteria`

src/tests/unit/platform/test_users.py
abowersox-cb: In my opinion, this adds _nothing_ to the tests, only increasing complexity for no good reason. It does not increase code coverage of the tests, for instance.  Nevertheless, I will do as you ask.

avanbrunt-cb: This is not intended to add anything to the tests from a functional stand point instead this maintains the perspective of how to use the SDK correctly. If we take short cuts and perform tests which will through an API error then our tests don't match to the functionality of the developers experience and is therefore an invalid testcase

avanbrunt-cb (approved):

PR #164, opened 2021-04-26
Enhance reputation_override_uat.py

gyanitsa-cb:
Fix reputation_override_uat.py to work as expected when there are additional IT_TOOL and CERT reputation overrides for the same organisation

## Pull request checklist

https://jira.carbonblack.local/browse/CBAPI-2542

## Pull Request Description

Currently the test works only on fresh environment. It fails if there are other reputation overrides of type IT_TOOLS or CERT in the organisation. The proposed solution resolves this issue.

- [ ] Yes
- [x] No

## How Has This Been Tested?

The test passes successfully when performed on non-fresh environment.

### DIFF
--- a/src/tests/uat/reputation_override_uat.py
+++ b/src/tests/uat/reputation_override_uat.py
@@ -26,11 +26,10 @@ Reputation Override:
 """
 
 # Standard library imports
+import pytest
 import sys
 
 # Internal library imports
-import pytest
-
 from cbc_sdk.helpers import build_cli_parser, get_cb_cloud_object
 from cbc_sdk.platform import ReputationOverride
 from cbc_sdk.errors import ObjectNotFoundError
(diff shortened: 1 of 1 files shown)


src/tests/uat/reputation_override_uat.py
```diff
@@ -132,19 +132,20 @@ def main():
...
-                     .set_override_type("IT_TOOL") \
-                     .sort_by("create_time", "asc")
-
-    assert len(delete_query) == 0
+    isDeleted = True
+    try:
+        cb.select(ReputationOverride, it_tool_override.id)
+        isDeleted = False
+    except ObjectNotFoundError:
+        pass
```
avanbrunt-cb, line 140: Instead of using a boolean I would recommend importing pytest and using the following code as this will fail if the code does not raise the expected exception
```
with pytest.raises(ObjectNotFoundError):
        cb.select(ReputationOverride, it_tool_override.id)
```

gyanitsa-cb: Done

emitreva-cb (approved):

avanbrunt-cb (approved):

PR #165, opened 2021-04-27
Fix sensor install on Compute Resource. Refactor Vulnerability models

avanbrunt-cb:
Refactor Vulnerability models and simplify structure

## Pull request checklist

N/A

## Pull Request Description

Fixed sensor install on Compute Resource
Refactored and simplified Vulnerability models

- [x] Yes
- [ ] No

## How Has This Been Tested?

Manual and CI/CD tests


abowersox-cb:
We stepped through this together during Tuesday's Kaffee Klatsch.

src/cbc_sdk/platform/devices.py
```diff
@@ -172,21 +171,11 @@ def update_sensor_version(self, sensor_version):
         """
         return self._cb.device_update_sensor_version([self._model_unique_id], sensor_version)
 
-    def vulnerability_refresh(self, vcenter_specific=False):
```
emitreva-cb, line 175: So we do not need to cover this API: https://developer.carbonblack.com/reference/carbon-black-cloud/workload-protection/latest/vulnerability-assessment/#perform-action-on-a-vcenter-device ?

src/cbc_sdk/workload/vulnerability_assessment.py
```diff
@@ -151,37 +52,26 @@ def _refresh(self):
...
-        resp = self._cb.get_object(self._build_api_request_uri())
-        resp['cve_id'] = self._model_unique_id
-        self._info = resp
-        self._last_refresh_time = time.time()
-        return True
-
-    def _build_api_request_uri(self, http_method="GET"):
-        """
-        Build the unique URL used to make requests for this object.
+        result = list(self._cb.select(Vulnerability).add_criteria("cve_id", self._model_unique_id))
```
emitreva-cb, line 55: I might not be reading this correctly, but to get a specific vulnerability we are using the search and not the : https://developer.carbonblack.com/reference/carbon-black-cloud/workload-protection/latest/vulnerability-assessment/#get-vulnerability-details in fact I think this API (Get Vulnerability Details) is not used at all.

src/cbc_sdk/platform/devices.py
avanbrunt-cb: The problem with this API endpoint is it requires the vm_id which is not available in the Device model so we are unable to perform this request. I've also found that there are no vcenter devices that don't have a device id so this API endpoint is mainly for the vSphere application to make API calls through the Appliance VM

src/cbc_sdk/workload/vulnerability_assessment.py
```diff
@@ -193,72 +83,161 @@ def _query_implementation(cls, cb, **kwargs):
...
-    swagger_meta_file = "workload/models/vulnerability.yaml"
-    url_additional = "devices/{}/vulnerabilities/summary"
+    class AssetView(list):
+        """Represents a list of Vulnerability for an organization."""
+        urlobject = "/vulnerability/assessment/api/v1/orgs/{}"
+        swagger_meta_file = "workload/models/vulnerabilityAssetView.yaml"
+
+        def __init__(self, cb, initial_data=None):
+            """
+            Initialize Vulnerability.OrgSummary object
```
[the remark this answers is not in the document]
avanbrunt-cb: Thanks good catch!

src/cbc_sdk/workload/vulnerability_assessment.py
avanbrunt-cb: If you look at the responses between the search and the vulnerability details, the search response contains the vulnerability details however it is `null` when the following is not included in the response `/_search?dataForExport=true`. So we can get both in a single call instead of needing to make 1 + n api calls when fetching the Vulnerabilities. The other aspect I noticed was the vulnerability details does not contain the necessary properties in order to make additional helper calls like get affected_assets as you need the os_product_id. 

The end result was finding that the _search request supported the cve_id property so we could get the exact cve vulnerability while keep the number of API calls and providing the entire vulnerability object

src/cbc_sdk/workload/vulnerability_assessment.py
```diff
@@ -24,125 +24,33 @@
...
-        return cb.get_object(req_url, query_params)
-
-
 class Vulnerability(NewBaseModel):
     """Represents a vulnerability"""
-    urlobject = '/vulnerability/assessment/api/v1'
-    urlobject_single = '/vulnerability/assessment/api/v1/vulnerabilities/{}'
+    urlobject = '/vulnerability/assessment/api/v1/orgs/{}'
     swagger_meta_file = "workload/models/vulnerability.yaml"
     url_additional = ""
```
emitreva-cb, line 31: I'm sorry I did not notice this earlier, but I think url_additional is not needed any more.

src/tests/uat/vulnerability_assessment_uat.py
```diff
@@ -233,17 +258,28 @@ def main():
...
+    else:
+        sdk_results = vulnerability._info
+
+    assert api_results == sdk_results, 'Test Failed Expected: {} Actual: {}'.format(api_results, sdk_results)
     print('Get vulnerability details for a specific CVE ID.....................OK')
 
-    api_results = get_affected_devices(data={'os_product_id': '313_0'})['results']
-    query = vulnerability.affected_assets('313_0')
+    api_results = get_affected_devices(cve_id=CVE_ID, data={'os_product_id': vulnerability.os_product_id})['results']
+    query = vulnerability.get_affected_assets()
```
emitreva-cb, line 279: When I ran the tests vulnerability is a list and as a result vulnerability.os_product_id and vulnerability.get_affected_assets() is raising an error AttributeError: 'list' object has no attribute 'xxx'

emitreva-cb (approved):
LGTM

src/cbc_sdk/workload/vulnerability_assessment.py
avanbrunt-cb: Done!

src/tests/uat/vulnerability_assessment_uat.py
avanbrunt-cb: I forgot that case in the follow up test. I've fixed it now

src/cbc_sdk/workload/vulnerability_assessment.py
```diff
@@ -151,37 +58,36 @@ def _refresh(self):
...
-        resp['cve_id'] = self._model_unique_id
-        self._info = resp
+        query = self._cb.select(Vulnerability).add_criteria("cve_id", self._model_unique_id)
+        results = list(query)
+
+        if len(results) > 1:
+            if self._info.get('os_product_id', None):
+                for vuln in results:
+                    if vuln.os_product_id == self._info["os_product_id"]:
+                        self._info = vuln._info
```
abowersox-cb, line 68: This loop here was marked as "untested" in Coveralls. Maybe you could slip in another test for this? I saw that test coverage on this file had gone _down._ It was still up above 94%, but....

abowersox-cb (commented):
We covered most of this in our "Bulgarian Holiday" Kaffee Klatsch review.  And you added the extra tests that I spotted the need for.

src/cbc_sdk/workload/vulnerability_assessment.py
avanbrunt-cb: Done

abowersox-cb (approved):

PR #167, opened 2021-04-28
Automate comparison of SDK and API calls results in the platform_devices_uat.py test

gyanitsa-cb:
## Pull request checklist

https://jira.carbonblack.local/browse/CBAPI-2540 

## Pull Request Description
Adding automated comparison of the results from the SDK and API calls in the platform_devices_uat.py test. Currently, the comparison should be performed manually by the user.

- [ ] Yes
- [x] No

## How Has This Been Tested?

When executed locally, the test passes successfully.

### DIFF
--- a/src/tests/uat/platform_devices_uat.py
+++ b/src/tests/uat/platform_devices_uat.py
@@ -11,8 +11,6 @@
 # * NON-INFRINGEMENT AND FITNESS FOR A PARTICULAR PURPOSE.
 
 """
-Validation is manual; either from information in the console or running an equivalent API call from postman
-
 To execute, a profile must be provided using the standard CBC Credentials.
 
 The following API calls are tested in this script:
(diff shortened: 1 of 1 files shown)


src/tests/uat/platform_devices_uat.py
```diff
@@ -24,37 +22,72 @@
...
 from cbc_sdk.helpers import build_cli_parser, get_cb_cloud_object
 from cbc_sdk.platform import Device
 
+HOSTNAME = ''
+ORG_KEY = ''
+HEADERS = {'X-Auth-Token': '', 'Content-Type': 'application/json'}
+SEARCH_DEVICES = '{}appservices/v6/orgs/{}/devices/_search'
+
+
+def search_devices_api_call():
```
avanbrunt-cb, line 38: Instead of using global variables to pass around the data would it make more sense to pass the `cb` variable in as a parameter then move the following code inside the function
```
    HEADERS['X-Auth-Token'] = cb.credentials.token
    ORG_KEY = cb.credentials.org_key
    HOSTNAME = cb.credentials.url
```

gyanitsa-cb: I am concerned that not using global variables would create code duplication in case we have several methods that perform API calls and need to invoke this information. Currently, this test has only one API call (respectively method) but thus we have it prepared to quickly add more.

avanbrunt-cb: That's reasonable. Given this is a UAT test that is fine however we would not want to use global variables in the core sdk code. 

For reference, there are no planned changes to the device APIs at the moment and the other endpoints are device actions. Here is the API documentation if we want to add those to this UAT test https://developer.carbonblack.com/reference/carbon-black-cloud/platform/latest/devices-api/#device-actions

avanbrunt-cb (approved):
Use of global variables is fine in UAT tests only

emitreva-cb (approved):

src/tests/uat/platform_devices_uat.py
gyanitsa-cb: I've added a Jira ticket to cover device actions.

Issue #162, opened 2021-04-23
[BUG]


82d28a:
**I am unable to set argument expiration_seconds to ...ubs.Binary.download_url because it is a property:**
 - OS: Windows
 - Carbon Black Cloud Products:  Enterprise EDR
 - Python Version: 3.8

**Describe the bug**
I am unable to set argument expiration_seconds to ...ubs.Binary.download_url because it is a property
When I pass in any argument I get ...TypeError: 'str' object is not callable

**Steps to Reproduce**
Steps to reproduce the behavior (Provide a log message if relevant):
file_pull = ubs.Binary(cb, sample_sha256_hash)
file_pull.download_url(expiration_seconds=30)

**Expected behavior**
download url string should be returned available for 30 seconds

**Screenshots**
none available

**Additional context**
no

avanbrunt-cb:
@82d28a We are going to pick up this bug in our next sprint

emitreva-cb:
This issue is fixed with the new release. Check : https://carbon-black-cloud-python-sdk.readthedocs.io/en/latest/unified-binary-store/

PR #169, opened 2021-05-11
Fixing issue 162 and one more bug and increase the test coverage

emitreva-cb:
## Pull request checklist

https://jira.carbonblack.local/browse/CBAPI-2567
https://github.com/carbonblack/carbon-black-cloud-sdk-python/issues/162

- [x] Yes
- [ ] No

download_url is no longer a property, but a method, so the proper way to get the dowload_url will be either file_pull.download_url(expiration_seconds=30) or file_pull.download_url()

## How Has This Been Tested?
With a python script.

### DIFF
--- a/docs/unified-binary-store.rst
+++ b/docs/unified-binary-store.rst
@@ -1,0 +1,54 @@
+Unified Binary Store
+====================
+
+The unified binary store (UBS) is a centralized service that is part of the Carbon Black Cloud. The UBS is responsible
+for storing all binaries and corresponding metadata for those binaries. The UBS comes packaged in with Enterprise EDR.
+
+Get Download URL
+----------------
+
+::
+
+    >>> from cbc_sdk import CBCloudAPI
+    >>> api = CBCloudAPI(profile='sample')
+    >>> from cbc_sdk.enterprise_edr.ubs import Binary
+    >>> ha256_hash = '8005557c1614c1e2c89f7db3702199de2b1e4605718fa32ff6ffdb2b41ed3759'
+    >>> binary = Binary(cb, ha256_hash)
+    >>> download_url = binary.download_url()
+
+*Note: This is going to generate download link for the binary that will be active for 1 hour (default expiration period).*
+
+Get Download URL Valid For Specific Period
+------------------------------------------
+
+We could set expiration period for the download link (in seconds).
+
+::
+
+    >>> from cbc_sdk import CBCloudAPI
+    >>> api = CBCloudAPI(profile='sample')
+    >>> from cbc_sdk.enterprise_edr.ubs import Binary
+    >>> ha256_hash = '8005557c1614c1e2c89f7db3702199de2b1e4605718fa32ff6ffdb2b41ed3759'
+    >>> binary = Binary(cb, ha256_hash)
+    >>> download_url = binary.download_url(expiration_seconds=30)
+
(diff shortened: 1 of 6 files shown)


avanbrunt-cb (approved):
Looks good. My one thought is we should add something to read the docs indicating that a `NonQueryableModel` error will still support the select by id but the search capability

emitreva-cb:
> Looks good. My one thought is we should add something to read the docs indicating that a `NonQueryableModel` error will still support the select by id but the search capability

Good point, added 

> Looks good. My one thought is we should add something to read the docs indicating that a `NonQueryableModel` error will still support the select by id but the search capability

Good point, I have included a guide how to obtain the download url with couple of ways to do it.

docs/unified-binary-store.rst
```diff
@@ -0,0 +1,54 @@
...
+
+Get Download URL
+----------------
+
+::
+
+    >>> from cbc_sdk import CBCloudAPI
+    >>> api = CBCloudAPI(profile='sample')
+    >>> from cbc_sdk.enterprise_edr.ubs import Binary
+    >>> ha256_hash = '8005557c1614c1e2c89f7db3702199de2b1e4605718fa32ff6ffdb2b41ed3759'
```
avanbrunt-cb, line 15: sha256_hash instead of ha256_hash

docs/unified-binary-store.rst
```diff
@@ -0,0 +1,54 @@
...
+    >>> binary = Binary(cb, ha256_hash)
+    >>> download_url = binary.download_url()
+
+*Note: This is going to generate download link for the binary that will be active for 1 hour (default expiration period).*
```
avanbrunt-cb, line 19: We should show an example value for what download_url will be

docs/unified-binary-store.rst
```diff
@@ -0,0 +1,54 @@
...
+
+    >>> from cbc_sdk import CBCloudAPI
+    >>> api = CBCloudAPI(profile='sample')
+    >>> from cbc_sdk.enterprise_edr.ubs import Binary
+    >>> ha256_hash = '8005557c1614c1e2c89f7db3702199de2b1e4605718fa32ff6ffdb2b41ed3759'
+    >>> binary = cb.select(Binary, ha256_hash)
+
+*Note: If we try to use* :code:`binary = cb.select(Binary)` *, it will fail with exception that the model is non queryable model.* 
+
+Find a full list of supported commands in the
```
avanbrunt-cb, line 52: I'm not sure what you mean by commands? Given this is the CBC SDK would it make more sense to reference the UBS portion of our Read the Docs? https://carbon-black-cloud-python-sdk.readthedocs.io/en/latest/cbc_sdk.enterprise_edr/#module-cbc_sdk.enterprise_edr.ubs

If there are additional API endpoints which we have not included in the CBC SDK then we need to create some stories and get them added

docs/unified-binary-store.rst
avanbrunt-cb: Apply in all locations

docs/unified-binary-store.rst
avanbrunt-cb: Apply in all locations

avanbrunt-cb (approved):

hkaragitliev-cb (approved):

lhilmes-cb (approved):

PR #170, opened 2021-05-12
CBAPI-2552: Fix bugs in helpers.py and increase test coverage

emitreva-cb:
## Pull request checklist

https://jira.carbonblack.local/browse/CBAPI-2552

- [ ] Yes
- [x] No

### DIFF
--- a/src/cbc_sdk/helpers.py
+++ b/src/cbc_sdk/helpers.py
@@ -92,7 +92,7 @@ def get_cb_cloud_object(args):
     return cb
 
 
-def get_object_by_name_or_id(cb, cls, name_field="name", id=None, name=None, force_init=True):
+def get_object_by_name_or_id(cb, cls, name_field="name", id=None, name='', force_init=True):
     """
     Locate an object in the API by either ID or name.
 
@@ -111,7 +111,7 @@ def get_object_by_name_or_id(cb, cls, name_field="name", id=None, name=None, for
     try:
         if id:
             attempted_to_find = "ID of {0:d}".format(id)
-            objs = [cb.select(cls, id, force_init=force_init)]
+            objs = [cb.select(cls, id)]
         else:
             attempted_to_find = "name {0:s}".format(name)
             objs = cb.select(cls).where("{0}:{1}".format(name_field, name))[::]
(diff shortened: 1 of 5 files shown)


src/cbc_sdk/helpers.py
```diff
@@ -92,7 +92,7 @@ def get_cb_cloud_object(args):
     return cb
 
 
-def get_object_by_name_or_id(cb, cls, name_field="name", id=None, name=None, force_init=True):
+def get_object_by_name_or_id(cb, cls, name_field="name", id=None, name='', force_init=True):
```
emitreva-cb, line 95: If name is not provided, then it fails to format it - attempted_to_find = "name {0:s}".format(name) and then fails with the error that  attempted_to_find is used before assignment.

src/cbc_sdk/helpers.py
```diff
@@ -111,7 +111,7 @@ def get_object_by_name_or_id(cb, cls, name_field="name", id=None, name=None, for
     try:
         if id:
             attempted_to_find = "ID of {0:d}".format(id)
-            objs = [cb.select(cls, id, force_init=force_init)]
```
emitreva-cb, line 114: As far as I saw NewBaseModel has force_init as parameter of the __init__, but not the classes that are inheriting that base class.

avanbrunt-cb (changes requested):
Can we also remove `parser.add_argument("--window", help="Define search window", default='3d')` from `build_cli_parser` as this argument is not needed for every parser. We will need to add it to any example scripts which use it but that should be a small change

src/cbc_sdk/helpers.py
```diff
@@ -111,7 +111,7 @@ def get_object_by_name_or_id(cb, cls, name_field="name", id=None, name=None, for
...
+            objs = [cb.select(cls, id)]
         else:
```
avanbrunt-cb, line 115: This function requires either an id or name we should through an error neither are provided. I would leave the `name=None` then convert the if else to an if elif else where the else raises an Exception indicating that either id or name should be provided. We should also consider the case where both are provided, do we want to fetch by id then verify if the name matches or raise an error before making an API request indicating only one can be specified

avanbrunt-cb (approved):
Thanks! Looks good

hkaragitliev-cb (approved):
LGTM

PR #172, opened 2021-05-17
CBAPI-2510: Add Permitted Roles API to SDK

abowersox-cb:
## Pull request checklist

https://jira.carbonblack.local/browse/CBAPI-2510

## Pull Request Description

Added the get_permitted_roles API to the Grant class. This allows the user to fetch the roles that they are permitted to manage.

- [ ] Yes
- [x] No

## How Has This Been Tested?

Unit test has been added for the new API. All added code is 100% covered. Coverage in grants.py remains at 99%.

### DIFF
--- a/src/cbc_sdk/platform/grants.py
+++ b/src/cbc_sdk/platform/grants.py
@@ -494,7 +494,7 @@ class Grant(MutableBaseModel):
             cb (CBCloudAPI): A reference to the CBCloudAPI object.
 
         Returns:
-            set: A set of string role URNs that we are permitted to manage (assign to users).
+            list: A list of string role URNs that we are permitted to manage (assign to users).
         """
         def _extractor(resp_data):
             """Pulls out all the lists of URN descriptors from the result."""
@@ -507,7 +507,7 @@ class Grant(MutableBaseModel):
                                                                                     token_split[1])
         data = list(_extractor(cb.get_object(url)))
         flat_data = [item for sublist in data for item in sublist]
-        return set([subitem['urn'] for subitem in flat_data if 'urn' in subitem])
+        return list(set([subitem['urn'] for subitem in flat_data if 'urn' in subitem]))
 
     @classmethod
     def create(cls, cb, template=None, **kwargs):
--- a/src/cbc_sdk/platform/grants.py
+++ b/src/cbc_sdk/platform/grants.py
@@ -486,16 +486,15 @@ class Grant(MutableBaseModel):
         self._refresh_if_needed(ret)
 
     @classmethod
-    def get_permitted_roles(cls, cb, userid):
+    def get_permitted_roles(cls, cb):
         """
-        Returns a list of all permitted roles that the specified user can have.
+        Returns a list of all permitted roles that we can assign to a user.
 
         Args:
             cb (CBCloudAPI): A reference to the CBCloudAPI object.
-            userid (int): ID of the user to get permitted roles for.
 
         Returns:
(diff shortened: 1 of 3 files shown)


src/cbc_sdk/platform/grants.py
```diff
@@ -485,6 +485,30 @@ def _delete_object(self):
...
+            set: A set of string role URNs that we are permitted to manage (assign to users).
+        """
+        def _extractor(resp_data):
+            """Pulls out all the lists of URN descriptors from the result."""
+            for d in resp_data.values():
+                for ll in d.values():
+                    yield ll
+
+        token_split = cb.credentials.token.split('/')
+        url = "/access/v3/orgs/{0}/principals/{1}/roles/permitted?type=USER".format(cb.credentials.org_key,
```
avanbrunt-cb, line 506: If we are sending the API_ID should `type=API_KEY`? Is it possible that I could get the permitted roles for a user by it's user_id while authenticating with an API_KEY?

avanbrunt-cb: Or does the type filter the roles by Access Level ones vs User roles?

avanbrunt-cb: If we can get the permitted roles for a User using an API_Key then we should consider adding this as a function to the User object as well

abowersox-cb: We can't get permitted roles for an arbitrary user. What we _can_ get are the set of roles that _you,_ the person making the request, are permitted to add to users.  Using USER as the type parameter returns this.  Using API_KEY returns the set of roles that you are permitted to add to API keys. This is not useful to us at this point.

avanbrunt-cb: To clarify, the `principal_id` in the API path has to match to the id of the principal that is authenticating otherwise we get a 403 Forbidden?

abowersox-cb: Yes. To be more precise, the principal_id in the path _has_ to be the "token" portion of the value you've supplied as the X-Auth-Token. So everything after the / character.  Otherwise, it's 403 Forbidden.

emitreva-cb (approved):
LGTM

src/cbc_sdk/platform/grants.py
```diff
@@ -486,27 +486,28 @@ def _delete_object(self):
...
                     yield ll
 
-        url = "/access/v3/orgs/{0}/principals/{1}/roles/permitted".format(cb.credentials.org_key, userid)
+        token_split = cb.credentials.token.split('/')
+        url = "/access/v3/orgs/{0}/principals/{1}/roles/permitted?type=USER".format(cb.credentials.org_key,
+                                                                                    token_split[1])
         data = list(_extractor(cb.get_object(url)))
         flat_data = [item for sublist in data for item in sublist]
-        return [subitem['urn'] for subitem in flat_data if 'urn' in subitem]
+        return set([subitem['urn'] for subitem in flat_data if 'urn' in subitem])
```
avanbrunt-cb, line 510: I'm not a fan of introducing a new type if we use `list` every where else

avanbrunt-cb: This also means the result is not subscriptable so they can use `result[0]`

avanbrunt-cb (approved):

src/cbc_sdk/platform/grants.py
abowersox-cb: Changed so it's a list.

PR #179, opened 2021-05-20
CBAPI-2597: Handle Macro-Operations for Users with No Grant

abowersox-cb:
## Pull request checklist

https://jira.carbonblack.local/browse/CBAPI-2597

## Pull Request Description

I found a serious flaw in a lot of the macro-operations. They don't take into account that users might not _have_ grants.  ( _None_ of the users I can access from dev01 has a grant.)  This adds some defensive code to manage those cases.  For most of these, "user has no grant" becomes a no-op. For some things, like "add profiles," adding them to a user with no grant _creates_ a grant with the profiles to be added.  The grant() operation also offers an optional parameter to create a new grant for a user if they don't have one already.

- [ ] Yes
- [x] No

## How Has This Been Tested?

Unit tests have been expanded to cover the no-grant cases, corresponding to the changes in the code.  Test coverage stands at 100% for users.py.

### DIFF
--- a/src/cbc_sdk/platform/users.py
+++ b/src/cbc_sdk/platform/users.py
@@ -251,12 +251,9 @@ class User(MutableBaseModel):
         """
         return f"psc:org:{self._cb.credentials.org_key}"
 
-    def grant(self, create=False):
+    def grant(self):
         """
         Locates the access grant for this user.
-
-        Args:
-            create (bool): If True and the user does not already have a grant, attempt to create one.
 
         Returns:
             Grant: Access grant for this user, or None if the user has none.
@@ -265,10 +262,6 @@ class User(MutableBaseModel):
         try:
             return query.one()
         except ObjectNotFoundError:
-            if create:
-                template = {'principal': self.urn, 'org_ref': self.org_urn, 'roles': [],
-                            'principal_name': f"{self.first_name} {self.last_name}", 'profiles': []}
-                return Grant.create(self._cb, template)
             return None
 
     @classmethod
--- a/src/cbc_sdk/platform/users.py
+++ b/src/cbc_sdk/platform/users.py
@@ -357,7 +357,7 @@ class User(MutableBaseModel):
             users (list[User]): List of User objects specifying users to be disabled.
         """
         for user in users:
-            user.disable_all_access()
+            user._disable_all_access()
 
     @classmethod
     def bulk_delete(cls, users):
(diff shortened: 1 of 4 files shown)


src/cbc_sdk/platform/users.py
```diff
@@ -251,15 +251,25 @@ def org_urn(self):
...
+            Grant: Access grant for this user, or None if the user has none.
         """
         query = self._cb.select(Grant).add_principal(self.urn, self.org_urn)
-        return query.one()
+        try:
+            return query.one()
+        except ObjectNotFoundError:
+            if create:
+                template = {'principal': self.urn, 'org_ref': self.org_urn, 'roles': [],
+                            'principal_name': f"{self.first_name} {self.last_name}", 'profiles': []}
```
avanbrunt-cb, line 270: Does this work? I recall someone mentioning at least roles or profiles was required

src/cbc_sdk/platform/users.py
```diff
@@ -370,10 +380,11 @@ def bulk_delete(cls, users):
     def disable_all_access(self):
         """Disables all access profiles held by ths user."""
         grant = self.grant()
-        for profile in grant.profiles_:
-            profile.set_disabled(True)
-            grant.touch()
-        grant.save()
+        if grant:
```
avanbrunt-cb, line 383: We should probably raise an error if there is no grant to disable so it doesn't appear to succeed when the legacy form is still being used

avanbrunt-cb: Apply to all other locations you added the grant conditional

src/cbc_sdk/platform/users.py
abowersox-cb: You know what? I can safely get rid of this. The one place I thought I'd need it, I wound up using a different mechanism. So I can just kill it here.

src/cbc_sdk/platform/users.py
abowersox-cb: Except for add_profiles, because there's a reasonable solution there (create a grant that has the profiles in it).

src/cbc_sdk/platform/users.py
```diff
@@ -408,17 +420,22 @@ def _internal_add_profiles(self, profile_templates):
...
+                for profile in grant.profiles_:
+                    if profile.matches_template(template) and profile.conditions['disabled']:
+                        profile.set_disabled(False)
+                        grant.touch()
+                        need_create = False
+                        break
+                if need_create:
+                    grant.create_profile(template)
+            grant.save()
+        else:
```
abowersox-cb, line 435: This is the one place where I don't need to throw an error if there's no grant, because it's reasonable to create a new grant containing the profiles in question.

src/cbc_sdk/platform/users.py
abowersox-cb: Addressed this in most recent commit.  Let me know what you think.

emitreva-cb (approved):

avanbrunt-cb (approved):

PR #180, opened 2021-05-25
CBAPI-2606: Receive enum in upper or lower case

emitreva-cb:
## Pull request checklist

https://jira.carbonblack.local/browse/CBAPI-2606

- [ ] Yes
- [x] No


avanbrunt-cb (approved):

kebringer-cb (commented):
There is also a status of "CANCELLED".  Does this to be listed or handled specifically?
https://developer.carbonblack.com/reference/carbon-black-cloud/platform/latest/live-response-api/#cancel-command

abowersox-cb (commented):
Test cases should use a mixture of lower-case and upper-case statuses, to ensure that the code will work even after the underlying endpoint is fixed.

mrangelov-cb (approved):

abowersox-cb (approved):
Reviewed during Kaffee Klatsch and feedback was applied.

emitreva-cb:
> There is also a status of "CANCELLED". Does this to be listed or handled specifically?
> https://developer.carbonblack.com/reference/carbon-black-cloud/platform/latest/live-response-api/#cancel-command

It doesn't seem to be used anywhere in the live response in the SDK.

avanbrunt-cb:
> > There is also a status of "CANCELLED". Does this to be listed or handled specifically?
> > https://developer.carbonblack.com/reference/carbon-black-cloud/platform/latest/live-response-api/#cancel-command
> 
> It doesn't seem to be used anywhere in the live response in the SDK.

You would use this in a PUT to the command to change the status to CANCELLED in order to terminate a command early. We might want to consider adding a terminate or cancel command function

PR #184, opened 2021-06-01
CBAPI-2353: User Mgmt, Grants and Access Profiles - User Tests

abowersox-cb:
## Pull request checklist

https://jira.carbonblack.local/browse/CBAPI-2353

## Pull Request Description

Added test scripts to validate functionality of the user and grant APIs, which also serve as examples of the functionality.  Also applied fixes to SDK code as necessary to reflect reality of interacting with the server.

- [ ] Yes
- [x] No

## How Has This Been Tested?

Scripts have been run against org on dev01.  Unit tests have been adjusted to reflect changes made for bug fix purposes in the SDK.


src/cbc_sdk/rest_api.py
```diff
@@ -59,6 +59,16 @@ def _perform_query(self, cls, **kwargs):
         else:
             raise ApiError("All Carbon Black Cloud models must provide _query_implementation")
 
+    @property
+    def org_urn(self):
+        """
+        Returns the URN of the current organization.
```
avanbrunt-cb, line 65: I would modify this docstring to be `Returns the URN based on the configured org_key.`

examples/platform/change_role.py
```diff
@@ -0,0 +1,45 @@
...
+from cbc_sdk.helpers import build_cli_parser, get_cb_cloud_object
+from cbc_sdk.platform import User
+import sys
+
+
+def main():
+    """Main function for the role changing script"""
+    parser = build_cli_parser("Change User's Role")
+    parser.add_argument('email', help='E-mail address of the user to change role of')
+    parser.add_argument('role', help='New role URN for the user')
```
avanbrunt-cb, line 24: Would you consider making this an input based script after the initial args to select one of the permitted roles?

avanbrunt-cb: Likely something to consider for the other example scripts too if the role argument is not provided

examples/platform/find_users_with_grants.py
```diff
@@ -0,0 +1,58 @@
...
+# * NON-INFRINGEMENT AND FITNESS FOR A PARTICULAR PURPOSE.
+
+"""Find users that have "interesting" grants (with profiles) and list them."""
+
+from cbc_sdk.helpers import build_cli_parser, get_cb_cloud_object
+from cbc_sdk.platform import User
+import sys
+
+
+def is_interesting(user):
```
avanbrunt-cb, line 20: This should be `has_grant` as `is_interesting` is not descriptive

examples/platform/list_permitted_roles.py
```diff
@@ -0,0 +1,33 @@
...
+# SPDX-License-Identifier: MIT
+# *******************************************************
+# *
+# * DISCLAIMER. THIS PROGRAM IS PROVIDED TO YOU "AS IS" WITHOUT
+# * WARRANTIES OR CONDITIONS OF ANY KIND, WHETHER ORAL OR WRITTEN,
+# * EXPRESS OR IMPLIED. THE AUTHOR SPECIFICALLY DISCLAIMS ANY IMPLIED
+# * WARRANTIES OR CONDITIONS OF MERCHANTABILITY, SATISFACTORY QUALITY,
+# * NON-INFRINGEMENT AND FITNESS FOR A PARTICULAR PURPOSE.
+
+"""List all the roles we are permitted to assign to users."""
```
avanbrunt-cb, line 13: Instead of `we are` it should be `the api_key is`

src/cbc_sdk/platform/users.py
```diff
@@ -149,13 +162,8 @@ def add_grant_profile(self, orgs, roles):
...
         def build(self):
-            """
-            Builds the new user and returns it.
-
-            Returns:
-                User: The new user object from the server.
-            """
-            return User._create_user(self._cb, self._creation_data)
+            """Builds the new user."""
+            User._create_user(self._cb, self._creation_data)
```
avanbrunt-cb, line 166: You added a comment below the query implementation I would add something similar to the docstring for context 

```
# N.B.: new user is not "findable" until activated and initial password set
```

examples/platform/name_a_user.py
```diff
@@ -0,0 +1,44 @@
...
+# SPDX-License-Identifier: MIT
+# *******************************************************
+# *
+# * DISCLAIMER. THIS PROGRAM IS PROVIDED TO YOU "AS IS" WITHOUT
+# * WARRANTIES OR CONDITIONS OF ANY KIND, WHETHER ORAL OR WRITTEN,
+# * EXPRESS OR IMPLIED. THE AUTHOR SPECIFICALLY DISCLAIMS ANY IMPLIED
+# * WARRANTIES OR CONDITIONS OF MERCHANTABILITY, SATISFACTORY QUALITY,
+# * NON-INFRINGEMENT AND FITNESS FOR A PARTICULAR PURPOSE.
+
+"""Example script which finds a user without a name and gives them one."""
```
avanbrunt-cb, line 13: Instead of finding a random user which is not a practical customer usecase this should update an existing user by email either their name or phone number. The code is also really dense and compressed which makes reading and comprehending the script a challenge for newer python developers.

examples/platform/set_user_enable.py
```diff
@@ -0,0 +1,60 @@
...
+import sys
+
+
+def peek_status(user):
+    """Returns True if user is enabled, False if disabled."""
+    grant = user.grant()
+    return not grant.revoked
+
+
+def poke_status(user, value):
```
avanbrunt-cb, line 26: This should be `get_status` and `update_status` as the function names are not intuitive

examples/platform/set_user_enable.py
```diff
@@ -0,0 +1,60 @@
...
+    if args.enable:
+        poke_status(user, True)
+        status = True
+    elif args.disable:
+        poke_status(user, False)
+        status = False
+    else:
+        status = peek_status(user)
+    word_status = "ENABLED" if status else "DISABLED"
+    print(f"Current status: {word_status}")
```
avanbrunt-cb, line 55: You can combine these lines into a single line

```
print(f"Current status: {'ENABLED' if status else 'DISABLED'}")
```

examples/platform/find_users_with_grants.py
abowersox-cb: Refactor -> Rename :)

examples/platform/list_permitted_roles.py
abowersox-cb: Altered

examples/platform/name_a_user.py
abowersox-cb: I may just dump this whole script as it doesn't add much useful to the examples. Would that be OK?

examples/platform/set_user_enable.py
abowersox-cb: You never coded in BASIC on the old late-70's micros, did you? In any event, two instances of Refactor -> Rename. :)

examples/platform/set_user_enable.py
abowersox-cb: OK, I'll do that!

src/cbc_sdk/platform/users.py
abowersox-cb: I fleshed the docstring out a bit along these lines.

src/cbc_sdk/rest_api.py
abowersox-cb: So modified.

abowersox-cb (commented):
Updates made based on above.

src/cbc_sdk/platform/users.py
avanbrunt-cb: Can we add the same note to the `def build` function as well

examples/platform/name_a_user.py
avanbrunt-cb: That works too

examples/platform/change_role.py
avanbrunt-cb: What I was thinking was that the role argument would be optional so that if it is not specified then you would print the the permitted roles for the API key so that the user can choose one of them. I really like the concept of this script however a user may not know the role urns so being able to let them see and choose would be helpful. By making the role argument optional a user could also provide the property if they already know the URN

```
What role would you like to configure?
1. View Only
2. Security Analyst I
3. Security Analyst II
```

src/cbc_sdk/platform/users.py
abowersox-cb: Copy-pasta done.

examples/platform/name_a_user.py
abowersox-cb: It's been sent to the bit bucket.

examples/platform/change_role.py
abowersox-cb: `Grant.get_permitted_roles` only returns the URNs of the roles, not any human-readable text such as you want to display here.  We'd have to change that API call if you wanted to do this.

avanbrunt-cb: There are other useful properties in the roles model, we should consider how to best to expose that information in the CBC SDK. Would it make sense to rename `get_permitted_roles` to `get_permitted_role_urns` and create an additional function to return the entire role object? Or is there another better option to provide the full role context?

avanbrunt-cb: https://developer.carbonblack.com/reference/carbon-black-cloud/platform/latest/access-profiles-and-grants/#get-permitted-roles

abowersox-cb: Well, we could go _completely_ down the rabbit hole here and make Role a first-class object within the SDK that can be queried via the `cb.select()` mechanism, but do you want to delay the 1.3 release of the SDK long enough to do this?  To put it another way, what's the _simplest_ thing we can do to satisfy your concerns on this particular item?

avanbrunt-cb (approved):

examples/platform/change_role.py
avanbrunt-cb: Capturing the effort to better handle roles in a future story

abowersox-cb: Following discussion during sprint planning, I have renamed `get_permitted_roles` to `get_permitted_role_urns` to leave room for future expansion of the role information we return, by whatever mechanism.  It's not worth holding up the 1.3 release to do anything else right now.

examples/platform/set_user_enable.py
```diff
@@ -0,0 +1,59 @@
...
+    grant.revoked = not value
+    grant.save()
+
+
+def main():
+    """Main function for the role changing script"""
+    parser = build_cli_parser("Set User Enable/Disable")
+    parser.add_argument('email', help='E-mail address of the user to set status of')
+    parser.add_argument('-e', '--enable', action='store_true', help='Enable this user')
+    parser.add_argument('-d', '--disable', action='store_true', help='Disable this user')
```
emitreva-cb, line 40: The only thing I was thinking is whether there should be some validations that -e and -d are not both present, but considering that only enabled would be applied maybe we do not need to worry.

emitreva-cb (approved):
LGTM

examples/platform/set_user_enable.py
abowersox-cb: Probably not. This is just an example and everyone can see the control flow favors "enable" over "disable."

PR #198, opened 2021-06-23
CBC SDK to use futures to make Live Response Async better

emitreva-cb:
## Pull request checklist

- [ ] Yes
- [ ] No


hkaragitliev-cb:
Wondering whether we should move live_response_api.py under platform?

emitreva-cb:
> Wondering whether we should move live_response_api.py under platform?

Seems like a good suggestion. I am not sure about the initial reasoning for including it outside of platform. If there are no objection I am going to move it under platforms.

avanbrunt-cb:
> > Wondering whether we should move live_response_api.py under platform?
> 
> Seems like a good suggestion. I am not sure about the initial reasoning for including it outside of platform. If there are no objection I am going to move it under platform.

This is a breaking change, do not make this change

emitreva-cb:
> > > Wondering whether we should move live_response_api.py under platform?
> > 
> > 
> > Seems like a good suggestion. I am not sure about the initial reasoning for including it outside of platform. If there are no objection I am going to move it under platform.
> 
> This is a breaking change, do not make this change

OK.

avanbrunt-cb:
> > > > Wondering whether we should move live_response_api.py under platform?
> > > 
> > > 
> > > Seems like a good suggestion. I am not sure about the initial reasoning for including it outside of platform. If there are no objection I am going to move it under platform.
> > 
> > 
> > This is a breaking change, do not make this change
> 
> OK.

If this is something we want to consider then we need to ensure the imports are not broken by making the live_response_api python file/package available in original location so we don't break this import statement `from cbc_sdk.live_response_api`

src/cbc_sdk/live_response_api.py
```diff
@@ -1032,6 +1210,8 @@ def __init__(self, cb, timeout=30, keepalive_sessions=False):
         self._keepalive_sessions = keepalive_sessions
         self._init_poll_delay = 1
         self._init_poll_timeout = 360
+        self._async_executor = None
+        self._thread_pool_count = self.THREAD_POOL_COUNT
```
avanbrunt-cb, line 1214: This should be configurable in the `__init__` function

src/cbc_sdk/live_response_api.py
```diff
@@ -1100,7 +1296,21 @@ def stop_keepalive_thread(self):
...
+    def _return_existing_session(self, device_id, session):
+        self._sessions[device_id]._refcount += 1
+        return session
+
+    def _get_session_obj(self, device_id, session_id):
+        session_id, session_data = self._wait_create_session(device_id, session_id)
+        session = self.cblr_session_cls(self, session_id, device_id, session_data=session_data)
+        return session
+
+    def _create_and_store_session(self, device_id, session_id):
```
avanbrunt-cb, line 1308: These don't need exposed and can be created as inside functions on `request_session`

src/tests/uat/live_response_api_uat.py
```diff
@@ -89,23 +91,9 @@ def flatten_processes_response_filter(response, process_name):
...
-    if print_detail:
-        print(f"args provided {args}")
-
-    cb = get_cb_cloud_object(args)
-    HEADERS['X-Auth-Token'] = cb.credentials.token
-    ORG_KEY = cb.credentials.org_key
-    HOSTNAME = cb.credentials.url
-
-    print(13 * ' ' + 'Live Response Commands')
+def synchronous_version(cb):
```
avanbrunt-cb, line 94: I would split this into two separate UAT tests as this is a long file now and is hard to read and understand

src/tests/unit/test_live_response_api.py
```diff
@@ -384,6 +469,111 @@ def test_get_file(cbcsdk_mock, connection_mock):
...
+    cbcsdk_mock.mock_request('GET', '/appservices/v6/orgs/test/liveresponse/sessions/1:2468', SESSION_POLL_RESP)
+    cbcsdk_mock.mock_request('GET', '/appservices/v6/orgs/test/devices/2468', DEVICE_RESPONSE)
+    cbcsdk_mock.mock_request('GET', '/appservices/v6/orgs/test/liveresponse/sessions/1:2468/commands/7',
+                             GET_FILE_COMMAND_RESP)
+    cbcsdk_mock.mock_request('DELETE', '/appservices/v6/orgs/test/liveresponse/sessions/1:2468', None)
+    cbcsdk_mock.mock_request('DELETE', '/appservices/v6/orgs/test/liveresponse/sessions/1:2468/commands/7', None)
+    manager = LiveResponseSessionManager(cbcsdk_mock.api)
+    with manager.request_session(2468) as session:
+        try:
+            session.cancel_command(7)
```
avanbrunt-cb, line 572: This needs an assert to verify successful completion of the command. Currently this only verifies that an exception did not occur

src/tests/unit/test_live_response_api.py
```diff
@@ -384,6 +469,111 @@ def test_get_file(cbcsdk_mock, connection_mock):
...
+                             GET_FILE_END_RESP)
+    cbcsdk_mock.mock_request('DELETE', '/appservices/v6/orgs/test/liveresponse/sessions/1:2468', None)
+    manager = LiveResponseSessionManager(cbcsdk_mock.api)
+    with manager.request_session(2468) as session:
+        with pytest.raises(ApiError) as excinfo:
+            session.cancel_command(7)
+        assert excinfo.value.__str__().startswith('Cannot cancel command in status COMPLETE')
+
+
+def test_cancel_pending_command(cbcsdk_mock):
```
avanbrunt-cb, line 560: We need a test to verify the logic flow for when a user cancels a command and what happens if a user then does a `command.results()`. will quickly return None? or some other result? or will it hang?

Do we also have a sample if an exception is raised during the async loop? How does the logic differ for each location we run the async loop? Can the exception handling be managed easily for all commands or do we need a test for each unique case?

emitreva-cb: With the previous implementation of poll_status, indeed if we cancel a command and then try to get `command.result()` it will hang, but I have added an extra if statement in poll_status to check for status cancelled and if the status is cancelled to raise ApiError (do not know whether this error is appropriate).

For the second part I do not follow completely, but if an exception is to be raised from async operation, it will be raised once we try to get the result with result(), it will not be raised before that even if there is something wrong.

src/cbc_sdk/live_response_api.py
```diff
@@ -107,6 +105,8 @@ def __init__(self, cblr_manager, session_id, device_id, session_data=None):
         self.device_id = device_id
         self._cblr_manager = cblr_manager
         self._cb = cblr_manager._cb
+        self._async_executor = None
+        self._thread_pool_count = self.THREAD_POOL_COUNT
```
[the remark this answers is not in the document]
emitreva-cb: Added as optional parameter.

src/cbc_sdk/live_response_api.py
```diff
@@ -363,29 +460,48 @@ def walk(self, top, topdown=True, onerror=None, followlinks=False):
...
+                                                                                        timeout=10,
+                                                                                        delay=0.1))
         try:
             self._poll_command(command_id, timeout=10, delay=0.1)
         except TimeoutError:
             return False
 
         return True
 
+    def _wait_for_completion(self, command_id, wait_for_completion, wait_for_output, wait_timeout, data):
```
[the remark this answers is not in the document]
emitreva-cb: Done.

src/cbc_sdk/live_response_api.py
```diff
@@ -513,36 +636,50 @@ def list_registry_keys_and_values(self, regkey):
...
                 containing value data, name, and type).
         """
         data = {"name": "reg enum key", "path": regkey}
         resp = self._lr_post_command(data).json()
         command_id = resp.get('id')
+        if async_mode:
+            return command_id, self._async_submit(lambda arg, kwarg: self._list_registry_keys_and_values(command_id))
+        return self._list_registry_keys_and_values(command_id)
+
+    def _list_registry_keys_and_values(self, command_id):
```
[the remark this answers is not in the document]
emitreva-cb: Done.

src/cbc_sdk/live_response_api.py
emitreva-cb: Added as optional parameter.

src/cbc_sdk/live_response_api.py
emitreva-cb: Done.

src/tests/uat/live_response_api_uat.py
emitreva-cb: Done.

src/tests/unit/test_live_response_api.py
emitreva-cb: Done.

src/tests/unit/test_live_response_api.py
avanbrunt-cb: Cool! That makes sense. Do we have a test for that as it's not clear if what you mentioned is covered in a unit test

emitreva-cb: Yep, but I have used only one command for that,  the function is `test_get_file_cancelled`.

avanbrunt-cb: That looks to cover it from the API side but it doesn't seem to account for the async case. I would expect a test that creates the command and returns a future. Then cancel the command and wait for the result

avanbrunt-cb (approved):

hkaragitliev-cb (approved):

PR #202, opened 2021-08-02
CBAPI-2716 & CBAPI-2769: Add Watchlist Utility / Macro functions to SDK

abowersox-cb:
## Pull request checklist

https://jira.carbonblack.local/browse/CBAPI-2716
https://jira.carbonblack.local/browse/CBAPI-2769

## Pull Request Description

Add additional utility/macro functions to watchlists, feeds, reports, and IOCs to simplify certain operations.

- [ ] Yes
- [x] No

## How Has This Been Tested?

Unit tests have been added for all new code, and, in many cases, for existing code in the threat_intelligence.py file as well. Test coverage of that file has gone from 61% to 94%.


src/cbc_sdk/enterprise_edr/threat_intelligence.py
```diff
@@ -482,14 +960,166 @@ def __init__(self, cb, model_unique_id=None, initial_data=None,
...
+            """
+            Initialize a new ReportBuilder.
+
+            Args:
+                cb (CBCloudAPI): A reference to the CBCloudAPI object.
+                report_body (dict): Partial report body which should be filled in with all "required" fields.
+            """
+            self._cb = cb
+            self._report_body = report_body
+            self._report_body['tags'] = []
```
avanbrunt-cb, line 977: What if report_body contains tags? This is going to overwrite the key

src/cbc_sdk/enterprise_edr/threat_intelligence.py
```diff
@@ -482,14 +960,166 @@ def __init__(self, cb, model_unique_id=None, initial_data=None,
...
+            Returns:
+                Report: The new Report.
+            """
+            report = Report(self._cb, None, self._report_body)
+            report._iocs_v2 = self._iocs
+            report._iocs_v2_need_sync = True
+            return report
+
+    @classmethod
+    def create(cls, cb, title, description, severity, timestamp=None):
```
avanbrunt-cb, line 1097: This should expose tags as a list of strings to be set in the constructor. You can default to an empty array

src/cbc_sdk/enterprise_edr/threat_intelligence.py
```diff
@@ -509,18 +1139,28 @@ def save_watchlist(self):
         return self
 
     def validate(self):
-        """Validates this report's state.
+        """
+        Validates this report's state.
```
avanbrunt-cb, line 1143: What is a report state? Are we validating that the report contents are supported? A more descriptive docstring is needed

src/cbc_sdk/enterprise_edr/threat_intelligence.py
```diff
@@ -509,18 +1139,28 @@ def save_watchlist(self):
...
 
         Raises:
-            InvalidObjectError: If the report's state is invalid
+            InvalidObjectError: If the report's state is invalid.
         """
+        id_substitute = False
```
avanbrunt-cb, line 1148: What is the purpose to `id_substitute`?

src/cbc_sdk/enterprise_edr/threat_intelligence.py
```diff
@@ -558,14 +1197,18 @@ def update(self, **kwargs):
...
         for key, value in kwargs.items():
             if key in self._info:
                 self._info[key] = value
+            if key == 'iocs_v2':
+                iocs_v2_poked = True
 
         if self.iocs:
             self._iocs = IOC(self._cb, initial_data=self.iocs, report_id=self.id)
-        if self.iocs_v2:
+        if self.iocs_v2 and iocs_v2_poked:
```
avanbrunt-cb, line 1209: Don't create a variable when you can check `'iocs_v2' in kwargs`

src/cbc_sdk/enterprise_edr/threat_intelligence.py
```diff
@@ -509,18 +1139,28 @@ def save_watchlist(self):
...
         super(Report, self).validate()
+        if id_substitute:
+            del self._info['id']
 
         if self.link and not validators.url(self.link):
             raise InvalidObjectError("link should be a valid URL")
 
         if self.iocs_v2:
             [ioc.validate() for ioc in self._iocs_v2]
+        if self._iocs_v2_need_sync:
```
avanbrunt-cb, line 1161: This is a confusing variable name as it doesn't need sync'd instead there are local changes which need pushed. I interpret need_sync as there are new changes on the CBC backend which need pulled.

avanbrunt-cb: Consider using a name schema like dirty as we use that already in https://github.com/carbonblack/carbon-black-cloud-sdk-python/blob/develop/src/cbc_sdk/base.py#L376

avanbrunt-cb (commented):
Overall great changes!

docs/watchlists-feeds-reports.rst
```diff
@@ -48,25 +48,28 @@ In this example, a report is created, adding one or more IOCs to it:
...
 
 Alternatively, you can update an existing report, adding more IOCs and/or replacing existing ones.  To find an existing
-report associated with a watchlist, you must look in the watchlist's ``reports`` collection.
+report associated with a watchlist, you must look in the watchlist's ``reports`` collection:
+
+::
+
+    >>> from cbc_sdk.enterprise_edr import Watchlist, Report
+    >>> watchlist = api.select('Watchlist', 'R4cMgFIhRaakgk749MRr6Q')
+    >>> rptlist = [report for report in watchlist.reports where report.id == '47474d40-1f94-4995-b6d9-1d1eea3528b3']
```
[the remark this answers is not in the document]
abowersox-cb: It was shortened to keep the line from exceeding maximum line length.

docs/watchlists-feeds-reports.rst
```diff
@@ -75,17 +78,18 @@ Now, add the new Report to a new Watchlist:
...
 
-**Note:** ``classifier`` *must* be ``None``, unless the new watchlist is being subscribed to a feed.
+If you already have an existing Watchlist you wish to enhance, you can add Reports to the existing Watchlist:
 
-If you already have an existing Watchlist you wish to enhance, you can add Reports to the existing Watchlist.
+::
+
+    >>> from cbc_sdk.enterprise_edr import Watchlist
+    >>> watchlist = api.select('Watchlist', 'R4cMgFIhRaakgk749MRr6Q')
+    >>> watchlist.add_reports([report])
```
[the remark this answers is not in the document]
abowersox-cb: Report objects.

docs/watchlists-feeds-reports.rst
```diff
@@ -227,16 +231,12 @@ A new feed may be created as follows (assuming the new report for that feed is s
...
-    ...                 'summary': 'Any signs of suspicious applications running on our endpoints',
-    ...                 'category': 'external_threat_intel',
-    ...                 'source_label': 'Where the info is coming from'},
-    ...                 'reports': [ {...}, {...} ]
-    ...             }
-    >>> feed = api.create(Feed, feed_data)
+    >>> builder = Feed.create(api, 'Suspicious Applications', 'http://example.com/location',
+    ...                       'Any signs of suspicious applications running on our endpoints', 'external_threat_intel')
+    >>> builder.set_source_label('Where the info is coming from')
+    >>> builder.add_reports([...])
```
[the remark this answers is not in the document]
abowersox-cb: Again, Report objects.

docs/watchlists-feeds-reports.rst
```diff
@@ -256,11 +256,7 @@ To subscribe to a feed, a new watchlist must be created around it:
 
 ::
 
-    >>> watchlist = Watchlist(api, None, {"alerts_enabled": True,
-    ...                                   "classifier": {"feed_id": feed.id},
-    ...                                   "description": "Subscription to the new feed",
-    ...                                   "name": "Subscribed feed",
-    ...                                   "tags_enabled": True})
+    >>> watchlist = Watchlist.create_from_feed(feed, "Subscribed feed", "Subscription to the new feed")
```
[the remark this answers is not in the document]
abowersox-cb: Not in create_from_feed.

src/cbc_sdk/enterprise_edr/threat_intelligence.py
```diff
@@ -14,6 +14,9 @@
 """Model Classes for Enterprise Endpoint Detection and Response"""
 
 from __future__ import absolute_import
+
+import uuid
```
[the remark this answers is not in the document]
abowersox-cb: This has not been confirmed.

src/cbc_sdk/enterprise_edr/threat_intelligence.py
```diff
@@ -64,6 +63,162 @@ def __init__(self, cb, model_unique_id=None, initial_data=None):
...
+
+        def set_name(self, name):
+            """
+            Sets the name for the new watchlist.
+
+            Args:
+                name (str): New name for the watchlist.
+
+            Returns:
+                FeedBuilder: This object.
```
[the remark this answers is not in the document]
abowersox-cb: Oof. That's what I get for doing to much copy-pasta. Fixing this.

src/cbc_sdk/enterprise_edr/threat_intelligence.py
```diff
@@ -64,6 +63,162 @@ def __init__(self, cb, model_unique_id=None, initial_data=None):
...
+            WatchlistBuilder: The builder for the new watchlist. Call build() to create the actual Watchlist.
+        """
+        return Watchlist.WatchlistBuilder(cb, name)
+
+    @classmethod
+    def create_from_feed(cls, feed, name=None, description=None):
+        """
+        Creates a new Watchlist that encapsulates a Feed.
+
+        Args:
```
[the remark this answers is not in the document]
abowersox-cb: Won't that add too many parameters to the parameters list? So if you wanted to just set alerts_enabled, you might have to do Watchlist.create_from_feed(feed, None, None, True) or something.

src/cbc_sdk/enterprise_edr/threat_intelligence.py
```diff
@@ -254,17 +417,57 @@ def reports(self):
...
     """Represents an Enterprise EDR feed's metadata."""
     urlobject = "/threathunter/feedmgr/v2/orgs/{}/feeds"
     urlobject_single = "/threathunter/feedmgr/v2/orgs/{}/feeds/{}"
     primary_key = "id"
     swagger_meta_file = "enterprise_edr/models/feed.yaml"
-
-    @classmethod
-    def _query_implementation(self, cb, **kwargs):
-        return FeedQuery(self, cb)
+    REPORT_VALIDATION = [
```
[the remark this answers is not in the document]
abowersox-cb: I don't know what "schemas" are. I just did what was necessary to try and keep the validation functions from getting out of hand.

src/cbc_sdk/enterprise_edr/threat_intelligence.py
```diff
@@ -431,16 +816,99 @@ def append_reports(self, reports):
         """
         if not self.id:
             raise InvalidObjectError("missing feed ID")
+        self._overwrite_reports(self._reports + reports, [])
 
-        rep_dicts = [report._info for report in reports]
-        rep_dicts += [report._info for report in self.reports]
-        body = {"reports": rep_dicts}
+    @classmethod
+    def _validate_against_criteria(cls, value_dict, criteria, tag):
```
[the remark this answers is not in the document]
abowersox-cb: Again, I don't know what "schemas" are. I just did what was necessary to try and keep the validation functions from getting out of hand.

src/cbc_sdk/enterprise_edr/threat_intelligence.py
abowersox-cb: report_body will never contain tags because it's constructed in Report.create() which takes discrete parameters and builds the dict itself.

src/cbc_sdk/enterprise_edr/threat_intelligence.py
abowersox-cb: Why, when it's just as easy to call add_tag on the builder before calling build?

src/cbc_sdk/enterprise_edr/threat_intelligence.py
abowersox-cb: I didn't change the actual docstring, just re-indented it. I don't know who wrote that originally.

src/cbc_sdk/enterprise_edr/threat_intelligence.py
abowersox-cb: `id_substitute` records whether we had to add a temporary id value before calling the superclass validate, and thus whether we need to _remove_ said temporary id value afterwards. Without this, calling the superclass validate will fail because there's no id in there until it's actually saved.

src/cbc_sdk/enterprise_edr/threat_intelligence.py
abowersox-cb: How do you mean? I need to record if the arguments changed iocs_v2, so that we can rebuild the _iocs_v2 array and flag that it no longer needs to be synced.

src/cbc_sdk/enterprise_edr/threat_intelligence.py
abowersox-cb: need_sync is kind of different from that.  The Report object actually contains _two_ representations of the current list of reports, the JSON representation in the info array and the list of Report objects.  That flag gets set to True when you add an object to the array, it records that the JSON needs to be rebuilt.

abowersox-cb: It also gets set when you _remove_ an object from the array, too. Either way.

docs/watchlists-feeds-reports.rst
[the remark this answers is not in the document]
abowersox-cb: I'm trying to clear this up.

docs/watchlists-feeds-reports.rst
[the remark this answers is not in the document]
abowersox-cb: I'm trying to clear this up.

docs/watchlists-feeds-reports.rst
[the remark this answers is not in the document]
abowersox-cb: Lengthening it put the end of the line literally _just_ at the edge. But not over it.

src/cbc_sdk/enterprise_edr/threat_intelligence.py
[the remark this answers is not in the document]
avanbrunt-cb: Ok. I can't approve until we have confirmation

src/cbc_sdk/enterprise_edr/threat_intelligence.py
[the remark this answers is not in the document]
avanbrunt-cb: I provided a link which creates the schemas we are using to do the same validation in our STIX/TAXII example.  Please review and give me your thoughts as the custom function and design will be harder to maintain than something we use else where in the repo

src/cbc_sdk/enterprise_edr/threat_intelligence.py
avanbrunt-cb: You assume that is the only place some will call this when you can create a builder directly if desired

src/cbc_sdk/enterprise_edr/threat_intelligence.py
avanbrunt-cb: Please write a better docstring for this function

src/cbc_sdk/enterprise_edr/threat_intelligence.py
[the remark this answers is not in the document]
avanbrunt-cb: I'm less concerned with the number of parameters when the function is a constructor or creating an object as the intention is to provide the ability to create the complete object. Given that this makes an API call and is not a builder class we should be exposing all the properties especially ones as important as the enable toggles. This also masks the properties such that a user will need to also look at the Watchlist model to see the additional properties that they need to set which is less likely to happen given people struggle at reading the docs in its entirety. 

```
watchlist = Watchlist.create_from_feed(feed, "Subscribed feed", "Subscription to the new feed")
watchlist.alerts_enabled = true
watchlist.tags_enabled = true
watchlist.save()
```

src/cbc_sdk/enterprise_edr/threat_intelligence.py
avanbrunt-cb: You don't need to create variable to check the kwargs if iocs_v2 exists in the for loop you can check the kwargs directly in the conditional see the following code

```
        for key, value in kwargs.items():
            if key in self._info:
                self._info[key] = value
        
        if self.iocs:
            self._iocs = IOC(self._cb, initial_data=self.iocs, report_id=self.id)
        if self.iocs_v2 and 'iocs_v2' in kwargs:
```

src/cbc_sdk/enterprise_edr/threat_intelligence.py
avanbrunt-cb: Looking over all the places we use `_iocs_v2_need_sync`. Why wait till the validate Reports to check the IOCs and push that into `_info` it seems like we should be doing that were we replace or add reports. How does this work if I update an IOC directly? What is the design for updating an existing IOC?

src/cbc_sdk/enterprise_edr/threat_intelligence.py
avanbrunt-cb: Given this a builder it's not going to increase the API calls. This should be fine since a user cans till create a report directly with the Report constructor with raw json if they don't want to use the builder

src/cbc_sdk/enterprise_edr/threat_intelligence.py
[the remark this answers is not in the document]
abowersox-cb: Added the additional parameters.

src/cbc_sdk/enterprise_edr/threat_intelligence.py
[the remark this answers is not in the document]
abowersox-cb: I see that file, but all it does is create Schema objects, it doesn't show how they are used. Could you provide a link to the documentation for the schema package so I at least have some clue how those objects work?

src/cbc_sdk/enterprise_edr/threat_intelligence.py
abowersox-cb: Why would anyone create a builder directly? The create() function encapsulates that process.  In the alternative, if they already have a fully-formatted dict, they can construct a Report directly.

Nonetheless, I put in a guard clause that will only initialize the tags if they aren't already there (or are there but aren't a list).

avanbrunt-cb: Thanks! I feel more comfortable with this implementation

src/cbc_sdk/enterprise_edr/threat_intelligence.py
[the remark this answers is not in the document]
avanbrunt-cb: Thanks!

src/cbc_sdk/enterprise_edr/threat_intelligence.py
abowersox-cb: Added it. You know, that was all supposed to make the process _easier,_ not _more_ complicated...

src/cbc_sdk/enterprise_edr/threat_intelligence.py
abowersox-cb: OK

src/cbc_sdk/enterprise_edr/threat_intelligence.py
avanbrunt-cb: It's an optional parameter and provides a clear indication of what is configurable. This is not more complicated. Thanks for adding the property.

src/cbc_sdk/enterprise_edr/threat_intelligence.py
abowersox-cb: OK

avanbrunt-cb: Thanks!

src/cbc_sdk/enterprise_edr/threat_intelligence.py
abowersox-cb: Because we don't want to rebuild that big list of dicts every time we do something to an IOC in the report. It wastes time and memory. We want to encourage the use of the objects whereever possible.  At the same time, we _also_ don't want to have to rebuild that big list of dicts if the IOCs haven't been touched, which, again, wastes time and memory.

src/cbc_sdk/enterprise_edr/threat_intelligence.py
avanbrunt-cb: Thanks! I am noticing the other validate functions for feed and IOC/v2 could use the same description. Would you update those too please?

src/cbc_sdk/enterprise_edr/threat_intelligence.py
avanbrunt-cb: Thanks for the explanation. I'm stuck on the name as it's not intuitive to me exactly what it represents though I do understand it better now. A small comment here https://github.com/carbonblack/carbon-black-cloud-sdk-python/blob/8df656cd6a540e4a181c899b6cdf5691569224b3/src/cbc_sdk/enterprise_edr/threat_intelligence.py#L965 will be enough to satisfy my confusion

src/cbc_sdk/enterprise_edr/threat_intelligence.py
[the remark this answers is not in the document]
abowersox-cb: That information was just what I needed.

The use of `schema` shrinks down the code a lot, but, oh, Lord, `schema` generates _UGLY_ error messages by default. I wound up having to override virtually every error message that could be thrown by the schema and tweak the way I get the message out of the `SchemaError` for the `InvalidObjectError`.  This is in keeping with your desire for "descriptive and intuitive" error handling.  It also makes the unit test function _way_ more complicated, or, rather, its `parametrize` blocks, but, after much effort, I got everything synced back up again.  I also had to add `schema` to the `requirements.txt` file; I hope that won't be a problem.

src/cbc_sdk/enterprise_edr/threat_intelligence.py
abowersox-cb: Every validate() function in this file now has a similar docstring.

src/cbc_sdk/enterprise_edr/threat_intelligence.py
abowersox-cb: I added the comment and also changed it to `_iocs_v2_need_rebuild`, as that's a better word for what it does (flags that we need to rebuild the `ioc_v2` element of `_info` based on what's in the `iocs_v2` array).

docs/watchlists-feeds-reports.rst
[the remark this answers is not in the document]
abowersox-cb: See if what's in there now looks better to you.

docs/watchlists-feeds-reports.rst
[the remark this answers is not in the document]
abowersox-cb: See if what's in there now looks better to you.

docs/watchlists-feeds-reports.rst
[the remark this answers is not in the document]
avanbrunt-cb: I would avoid using assume and state that intention so phrase it like so `# variable "report" is a Report object created from the previous example`

src/cbc_sdk/enterprise_edr/threat_intelligence.py
avanbrunt-cb: Thanks

src/cbc_sdk/enterprise_edr/threat_intelligence.py
avanbrunt-cb: Thanks! This is much clearer to me now

src/cbc_sdk/enterprise_edr/threat_intelligence.py
```diff
@@ -254,17 +421,66 @@ def reports(self):
...
     """Represents an Enterprise EDR feed's metadata."""
     urlobject = "/threathunter/feedmgr/v2/orgs/{}/feeds"
     urlobject_single = "/threathunter/feedmgr/v2/orgs/{}/feeds/{}"
     primary_key = "id"
     swagger_meta_file = "enterprise_edr/models/feed.yaml"
-
-    @classmethod
-    def _query_implementation(self, cb, **kwargs):
-        return FeedQuery(self, cb)
+    SCHEMA_IOCV2 = Schema(
```
avanbrunt-cb, line 455: Can we move these from `class Feed` to `class FeedModel`? The rest of the validation code can stay as is just the schemas in the base class makes more sense to me

src/cbc_sdk/enterprise_edr/threat_intelligence.py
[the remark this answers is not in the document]
avanbrunt-cb: Thanks! This will be a lot easier to maintain and expand in the future. The package add makes sense as we will probably use it in more places in the future

src/cbc_sdk/enterprise_edr/threat_intelligence.py
avanbrunt-cb: Looking into this I am a little confused why we are doing temporary substitution if these are required properties? If they are not required then we should remove them from the schema file here https://github.com/carbonblack/carbon-black-cloud-sdk-python/blob/develop/src/cbc_sdk/enterprise_edr/models/feed.yaml

A personal preference to reduce the code would be this alternative if the substitution is still needed then you don't need another variable since the string we set is a known no valid one. If this is needed for multiple properties then the cleanup can be a simple for loop through the keys though we should clean up the validation logic before adding substitutes or default to empty strings if a non None is needed

```
if not self._info.get('id', None):
    self._info['id'] = '**TEMP**'
super(Feed, self).validate()
if self._info['id'] === '**TEMP**':
    del self._info['id']
```

abowersox-cb: OK, this is the situation. The `id` value is _technically_ required, but only in `Reports` that have already been created on the server side.  A newly-built `Report` that has not yet saved to the server will not have an `id` yet, as the server assigns it.  However, both `update()` and `save_watchlist()` _call_ `validate()` (which is even more important now as it performs the rebuild of the `iocs_v2` element of `_info` as well), leading us to a chicken-and-egg problem. The `id` substitution lets us get around that problem and lets `validate()` complete OK so the `Report` can be saved.

src/cbc_sdk/enterprise_edr/threat_intelligence.py
abowersox-cb: Personally, I think it makes _less_ sense, but, if you insist, it shall be done.

docs/watchlists-feeds-reports.rst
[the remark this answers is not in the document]
abowersox-cb: I changed it, see what you think now.

src/cbc_sdk/enterprise_edr/threat_intelligence.py
avanbrunt-cb: Thanks! The reason is the Watchlist class inherits FeedModel not Feed so this is a future looking request in case we want to consume them there too

docs/watchlists-feeds-reports.rst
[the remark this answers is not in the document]
avanbrunt-cb: Looks great! Thanks!

src/cbc_sdk/enterprise_edr/threat_intelligence.py
avanbrunt-cb: So we can remove the other substitutes and just leave it with the id one?

abowersox-cb: Oh, you're not just looking at `Report`, are you? The fact that this was attached to `Report.validate()` was a bit misleading.

`Feed` requires `owner` and `access` as well. Those are set by the server, and are not required at create time.  So we have to fake those out in addition to `id`.

`Watchlist` requires `create_timestamp` and `last_update_timestamp`, which are also set by the server. At least there I _can_ flag the replacements with a "known not valid" value, namely -1.

avanbrunt-cb: That doesn't sound like requires as they are not required for creation they are created and managed by the backend which therefore makes them populated and always returned. If they are not required in the request then we don't need to validate them quite as strictly

abowersox-cb: I took the required flags out of the YAML and took the substitutions out of the code. The result passes the unit tests. Does that sit better with you?

avanbrunt-cb: As long as it functionally works against the CBC with those minimum fields. The required logic should match exactly what is required by the CBC backend. Have you run a manual test against the CBC backend?

abowersox-cb: No, I have not.

src/cbc_sdk/enterprise_edr/threat_intelligence.py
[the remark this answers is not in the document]
abowersox-cb: Transcript of session:

```
>>> from cbc_sdk import CBCloudAPI
>>> api = CBCloudAPI(profile='dev01')
>>> from cbc_sdk.enterprise_edr import Report, IOC_V2
>>> builder = Report.create(api, "Unsigned Browsers", "Unsigned processes impersonating browsers", 5)
>>> builder.add_tag("compliance").add_tag("unsigned_browsers")
<cbc_sdk.enterprise_edr.threat_intelligence.Report.ReportBuilder object at 0x0000011DAEAA8400>
>>> builder.add_ioc(IOC_V2.create_query(api, None, "process_name:chrome.exe NOT process_publisher_state:FILE_SIGNATURE_STATE_SIGNED"))
<cbc_sdk.enterprise_edr.threat_intelligence.Report.ReportBuilder object at 0x0000011DAEAA8400>
>>> report = builder.build()
Report created without feed ID or not from watchlist
>>> report._iocs_v2
[<cbc_sdk.enterprise_edr.threat_intelligence.IOC_V2: id ef4e68b6-c661-4d6e-9dc9-f845898a5ee8> @ https://defense-dev01.cbdtest.io]
>>> report.save_watchlist()
<cbc_sdk.enterprise_edr.threat_intelligence.Report: id rqjdfNcDTHqk7UuMOsJ9wg> @ https://defense-dev01.cbdtest.io
>>> report._iocs_v2
[<cbc_sdk.enterprise_edr.threat_intelligence.IOC_V2: id ef4e68b6-c661-4d6e-9dc9-f845898a5ee8> @ https://defense-dev01.cbdtest.io]
```
As you can see, the IOC ID is unchanged before and after the report is saved. This the only place where I use the UUID generator in this code.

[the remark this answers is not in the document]
abowersox-cb: Oddly enough, saving a Report for a Watchlist (with save_watchlist) gives it an ID, but saving it as part of a Feed does _not._ I modified `Feed.save()` to handle that little detail.  Transcript of the resulting code follows.

```
>>> from cbc_sdk import CBCloudAPI
>>> api = CBCloudAPI(profile='dev01')
>>> from cbc_sdk.enterprise_edr import Report, IOC_V2
>>> builder = Report.create(api, "Unsigned Browsers3", "Unsigned processes impersonating browsers", 6)
>>> builder.add_tag("compliance").add_tag("unsigned_browsers")
<cbc_sdk.enterprise_edr.threat_intelligence.Report.ReportBuilder object at 0x000001B2FC94D5B0>
>>> builder.add_ioc(IOC_V2.create_query(api, None, "process_name:opera3.exe NOT process_publisher_state:FILE_SIGNATURE_STATE_SIGNED"))
<cbc_sdk.enterprise_edr.threat_intelligence.Report.ReportBuilder object at 0x000001B2FC94D5B0>
>>> report = builder.build()
Report created without feed ID or not from watchlist
>>> from cbc_sdk.enterprise_edr import Feed
>>> builder = Feed.create(api, 'Suspicious Applications2', 'http://example.com/location', 'Any signs of suspicious applications running on our endpoints', 'external_threat_intel')
>>> builder.set_source_label('Where the info is coming from')
<cbc_sdk.enterprise_edr.threat_intelligence.Feed.FeedBuilder object at 0x000001B2FC94D640>
>>> builder.add_reports([report])
<cbc_sdk.enterprise_edr.threat_intelligence.Feed.FeedBuilder object at 0x000001B2FC94D640>
>>> feed = builder.build()
Report created without feed ID or not from watchlist
>>> feed.save()
<cbc_sdk.enterprise_edr.threat_intelligence.Feed: id zcA0XzYTXKprqKN3ZjDtw> @ https://defense-dev01.cbdtest.io
>>> feed._reports
[<cbc_sdk.enterprise_edr.threat_intelligence.Report: id 27eea3ee-ed67-4dce-ba32-9b7f52086620> @ https://defense-dev01.cbdtest.io]
```

src/cbc_sdk/enterprise_edr/threat_intelligence.py
abowersox-cb: I've proved out creating Report and Feed without those "missing" fields above.  See comment thread above.

avanbrunt-cb (approved):
Thanks for the verification and changes! This will be super useful for everyone

src/cbc_sdk/enterprise_edr/threat_intelligence.py
abowersox-cb: And Watchlist creation works, too. Transcript follows.

```
>>> from cbc_sdk import CBCloudAPI
>>> api = CBCloudAPI(profile='dev01')
>>> from cbc_sdk.enterprise_edr import Feed, Watchlist
>>> feed = api.select(Feed, 'zcA0XzYTXKprqKN3ZjDtw')
>>> watchlist = Watchlist.create_from_feed(feed)
>>> watchlist._info
{'name': 'Feed Suspicious Applications2', 'description': 'Any signs of suspicious applications running on our endpoints', 'tags_enabled': True, 'alerts_enabled': False, 'classifier': {'key': 'feed_id', 'value': 'zcA0XzYTXKprqKN3ZjDtw'}}
>>> watchlist.save()
<cbc_sdk.enterprise_edr.threat_intelligence.Watchlist: id Egg10PLQQJW54Mzs3cjt2Q> @ https://defense-dev01.cbdtest.io
>>> watchlist._info
{'name': 'Feed Suspicious Applications2', 'description': 'Any signs of suspicious applications running on our endpoints', 'tags_enabled': True, 'alerts_enabled': False, 'classifier': {'key': 'feed_id', 'value': 'zcA0XzYTXKprqKN3ZjDtw'}, 'id': 'Egg10PLQQJW54Mzs3cjt2Q', 'create_timestamp': 1628201980, 'last_update_timestamp': 1628201980, 'report_ids': None}
```

hkaragitliev-cb (approved):
Looking great, nice job, Amy!

PR #206, opened 2021-08-09
fix extra references for dependency on schema

abowersox-cb:
## Pull request checklist

## Pull Request Description

Fix the dependency on schema by adding it in two other places.

- [ ] Yes
- [ ] No


avanbrunt-cb (changes requested):
Make the change to develop since we have nothing else waiting there

PR #210, opened 2021-09-01
CBAPI-2380 - CBC SDK - add recommendation code

abowersox-cb:
## Pull request checklist

https://jira.carbonblack.local/browse/CBAPI-2380

## Pull Request Description

Add support for recommendation API to CBC SDK.

- [ ] Yes
- [x] No

## How Has This Been Tested?

Unit test coverage for newly added source is at 95%.

### DIFF
--- a/src/cbc_sdk/endpoint_standard/recommendation.py
+++ b/src/cbc_sdk/endpoint_standard/recommendation.py
@@ -1,0 +1,473 @@
+#!/usr/bin/env python3
+
+# *******************************************************
+# Copyright (c) VMware, Inc. 2020-2021. All Rights Reserved.
+# SPDX-License-Identifier: MIT
+# *******************************************************
+# *
+# * DISCLAIMER. THIS PROGRAM IS PROVIDED TO YOU "AS IS" WITHOUT
+# * WARRANTIES OR CONDITIONS OF ANY KIND, WHETHER ORAL OR WRITTEN,
+# * EXPRESS OR IMPLIED. THE AUTHOR SPECIFICALLY DISCLAIMS ANY IMPLIED
+# * WARRANTIES OR CONDITIONS OF MERCHANTABILITY, SATISFACTORY QUALITY,
+# * NON-INFRINGEMENT AND FITNESS FOR A PARTICULAR PURPOSE.
+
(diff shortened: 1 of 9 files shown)


src/cbc_sdk/endpoint_standard/recommendation.py
```diff
@@ -0,0 +1,473 @@
...
+    def __init__(self, cb, model_unique_id, initial_data=None):
+        """
+        Initialize the Recommendation object.
+
+        Args:
+            cb (BaseAPI): Reference to API object used to communicate with the server.
+            model_unique_id (str): ID of the recommendation represented.
+            initial_data (dict): Initial data used to populate the recommendation.
+        """
+        super(Recommendation, self).__init__(cb, model_unique_id, initial_data, full_doc=True)
```
emitreva-cb, line 45: If I saw correctly there is no /recommendation-service/v1/orgs/{org_key}/recommendation/{id}, but just search, so maybe we do not need model_unique_id. If you do not think that's correct just disregard.

abowersox-cb: It gets initialized when those objects get created in the query, with the `recommendation_id` member of the item data. And I _do_ need it when putting together the URL in the `_take_action` function.

src/cbc_sdk/endpoint_standard/recommendation.py
```diff
@@ -0,0 +1,473 @@
...
+        """
+        Reset the recommendation, undoing any created reputation override and setting it back to NEW state.
+
+        Args:
+            comment (str): Optional comment associated with the action.
+        """
+        self._take_action('RESET', comment)
+
+
+class RecommendationImpact(UnrefreshableModel):
```
avanbrunt-cb, line 136: Why make this a separate class instead of nesting inside the recommendation class?

avanbrunt-cb: Same for the other nested classes too

abowersox-cb: OK, I did that.

avanbrunt-cb (approved):
LGTM

PR #211, opened 2021-09-08
CBC SDK - add high-level tests (Recommendations)

abowersox-cb:
## Pull request checklist

https://jira.carbonblack.local/browse/CBAPI-2617

## Pull Request Description

UAT script for the Recommendations API

- [ ] Yes
- [x] No

## How Has This Been Tested?

This _is_ a test. It's been run against Dev01 successfully.

### DIFF
--- a/src/tests/uat/recommendation_uat.py
+++ b/src/tests/uat/recommendation_uat.py
@@ -1,0 +1,192 @@
+#!/usr/bin/env python
+# *******************************************************
+# Copyright (c) VMware, Inc. 2020-2021. All Rights Reserved.
+# SPDX-License-Identifier: MIT
+# *******************************************************
+# *
+# * DISCLAIMER. THIS PROGRAM IS PROVIDED TO YOU "AS IS" WITHOUT
+# * WARRANTIES OR CONDITIONS OF ANY KIND, WHETHER ORAL OR WRITTEN,
+# * EXPRESS OR IMPLIED. THE AUTHOR SPECIFICALLY DISCLAIMS ANY IMPLIED
+# * WARRANTIES OR CONDITIONS OF MERCHANTABILITY, SATISFACTORY QUALITY,
+# * NON-INFRINGEMENT AND FITNESS FOR A PARTICULAR PURPOSE.
+
+"""
+The following API calls are tested in this script.
+
+For the validation CBC API requests are used.
+
+To execute, a profile must be provided using the standard CBC Credentials.
+
+Recommendations:
+* Search Recommendations
+
+The following API calls were not covered:
+* Recommendation Workflow
+"""
+
+import sys
+import requests
+from cbc_sdk.helpers import build_cli_parser, get_cb_cloud_object
+from cbc_sdk.endpoint_standard import Recommendation
+
+
+def compare_impact(impact, raw_data):
+    """
+    Compare a RecommendationImpact object against the data that should represent it.
+
+    Args:
+        impact (RecommendationImpact): The RecommendationImpact object to compare.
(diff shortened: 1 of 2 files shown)


emitreva-cb (approved):

src/tests/uat/recommendation_uat.py
```diff
@@ -0,0 +1,192 @@
...
+
+def compare_impact(impact, raw_data):
+    """
+    Compare a RecommendationImpact object against the data that should represent it.
+
+    Args:
+        impact (RecommendationImpact): The RecommendationImpact object to compare.
+        raw_data (dict): The dictionary containing what should be the same data.
+    """
+    assert impact.event_count == raw_data['event_count']
```
gyanitsa-cb, line 41: Please, add error messages to be displayed in the console upon failure instead of a thrown error (for all asserts).

gyanitsa-cb:
Can we take actions on recommendations via the SDK? Like accepting, rejecting, ...

abowersox-cb:
> Can we take actions on recommendations via the SDK? Like accepting, rejecting

Yes. See the `accept`, `reject`, and `reset` methods on `Recommendation`.

> And can we create a recommendation via the SDK?

No. Recommendations are automatically created on the server.

> If yes, it's good to include a check for creating a recommendation and acting upon it.

We can't check the "recommendation workflow" API in a script because it doesn't return anything except a 204 No Content.

gyanitsa-cb:
Thanks, Amy! Great job :)

src/tests/uat/recommendation_uat.py
abowersox-cb: I had to rewrite most of the script, but I think I've given you what you want.

avanbrunt-cb:
@abowersox-cb The `ref_id` in the workflow sub resource for an ACCEPTED recommendation will point to the added reputation override created. This could be used to verify the recommendation was applied. The same could go for a RESET with the reputation override being deleted. As for REJECTED we could verify the workflow is updated.

See https://developer.carbonblack.com/reference/carbon-black-cloud/cb-defense/latest/recommendation-api/#workflow
```
Reference id for an ACCEPTED Recommendation which is the id of the created Reputation Override
```

abowersox-cb:
@avanbrunt-cb OK, but that's more complicated than a simple UAT script would suggest. Also, it requires that you actually _have_ an available recommendation on the server, and, this morning, as I was running the script, I couldn't find recommendations on _any_ environment I have access to. Even on DEV01, where I swear there were actual recommendations a week or two ago, because I used the output (as received via Postman) as a template for setting up my mock responses for unit tests.

avanbrunt-cb:
@abowersox-cb This is a good raise that we need a valid environment to test against. I would challenge that a `simple UAT script` should cover the entire flow which includes taking the actions and verifying they are completed correctly.

gyanitsa-cb:
I'm concerned about performing actions on a recommendation that we have not created with the test because we can't guarantee it won't affect another test running at the same time. And I totally agree that we need a valid environment that we can safely modify. I have some ideas in this direction that I want to research once we release the QRadar app.

src/tests/uat/recommendation_uat.py
gyanitsa-cb: The good thing now is that the test will go through all the checks before exiting but in case of failure it won't tell you which check failed which means failures will take more time to investigate. The initial way you proposed seems more readable. What do you think about using:
assert <expression>, <error_message>
It will end the test on the first failure but will give more user-friendly error and will be more readable.

abowersox-cb:
Withdrawing this pull request because this is going to require some serious rework to comply with everything that is wanted.

PR #212, opened 2021-09-16
CBAPI-2617: CBC SDK - add high-level tests (Recommendations)

abowersox-cb:
## Pull request checklist

https://jira.carbonblack.local/browse/CBAPI-2617

## Pull Request Description

High-level UAT tests for the Recommendations API, as well as other necessary changes to bring the code into line with reality.

- [ ] Yes
- [x] No

## How Has This Been Tested?

This _is_ a test script.  It has been run against dev01. Also, necessary unit tests have been added/altered where existing code was. Code coverage on the main source file is 95%.

### DIFF
--- a/src/tests/uat/recommendation_uat.py
+++ b/src/tests/uat/recommendation_uat.py
@@ -1,0 +1,192 @@
+#!/usr/bin/env python
+# *******************************************************
+# Copyright (c) VMware, Inc. 2020-2021. All Rights Reserved.
+# SPDX-License-Identifier: MIT
+# *******************************************************
+# *
+# * DISCLAIMER. THIS PROGRAM IS PROVIDED TO YOU "AS IS" WITHOUT
+# * WARRANTIES OR CONDITIONS OF ANY KIND, WHETHER ORAL OR WRITTEN,
+# * EXPRESS OR IMPLIED. THE AUTHOR SPECIFICALLY DISCLAIMS ANY IMPLIED
+# * WARRANTIES OR CONDITIONS OF MERCHANTABILITY, SATISFACTORY QUALITY,
+# * NON-INFRINGEMENT AND FITNESS FOR A PARTICULAR PURPOSE.
+
+"""
+The following API calls are tested in this script.
+
+For the validation CBC API requests are used.
+
+To execute, a profile must be provided using the standard CBC Credentials.
+
+Recommendations:
+* Search Recommendations
+
+The following API calls were not covered:
+* Recommendation Workflow
+"""
+
+import sys
+import requests
+from cbc_sdk.helpers import build_cli_parser, get_cb_cloud_object
+from cbc_sdk.endpoint_standard import Recommendation
+
+
+def compare_impact(impact, raw_data):
+    """
+    Compare a RecommendationImpact object against the data that should represent it.
+
+    Args:
+        impact (RecommendationImpact): The RecommendationImpact object to compare.
+        raw_data (dict): The dictionary containing what should be the same data.
+    """
+    assert impact.event_count == raw_data['event_count']
+    assert impact.impact_score == raw_data['impact_score']
+    assert impact.impacted_devices == raw_data['impacted_devices']
+    assert impact.org_adoption == raw_data['org_adoption']
+    assert impact.update_time == raw_data['update_time']
+
+
+def compare_application(application, raw_data):
+    """
+    Compare a RecommendationApplication object against the data that should represent it.
+
+    Args:
+        application (RecommendationApplication): The Recommendation object to compare.
+        raw_data (dict): The dictionary containing what should be the same data.
+    """
+    assert application.type == raw_data['type']
+    assert application.value == raw_data['value']
+
+
+def compare_new_rule(new_rule, raw_data):
+    """
+    Compare a RecommendationNewRule object against the data that should represent it.
+
+    Args:
+        new_rule (RecommendationNewRule): The RecommendationNewRule object to compare.
+        raw_data (dict): The dictionary containing what should be the same data.
+    """
+    assert new_rule.action == raw_data['action']
+    assert new_rule.certificate_authority == raw_data['certificate_authority']
+    assert new_rule.filename == raw_data['filename']
+    assert new_rule.include_child_processes == raw_data['include_child_processes']
+    assert new_rule.operation == raw_data['operation']
+    assert new_rule.override_list == raw_data['override_list']
+    assert new_rule.override_type == raw_data['override_type']
+    assert new_rule.path == raw_data['path']
+    assert new_rule.sha256_hash == raw_data['sha256_hash']
+    assert new_rule.signed_by == raw_data['signed_by']
+    if new_rule.application:
+        compare_application(new_rule.application, raw_data['application'])
+    else:
+        assert 'application' not in raw_data
+
+
+def compare_workflow(workflow, raw_data):
+    """
+    Compare a RecommendationWorkflow object against the data that should represent it.
+
+    Args:
+        workflow (RecommendationWorkflow): The RecommendationWorkflow object to compare.
+        raw_data (dict): The dictionary containing what should be the same data.
+    """
+    assert workflow.changed_by == raw_data['changed_by']
+    assert workflow.create_time == raw_data['create_time']
+    assert workflow.ref_id == raw_data['ref_id']
+    assert workflow.status == raw_data['status']
+    assert workflow.update_time == raw_data['update_time']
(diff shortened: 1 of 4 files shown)


abowersox-cb:
What's different from last time: The UAT now includes a full battery of tests for workflow. That meant that we had to do several things to the Recommendation object:

* Make it inherit from NewBaseModel instead of UnrefreshableModel, so it can be refreshed
* Implement `_refresh()` on Recommendation
* Make `_take_action()` call `_refresh()` after it performs a workflow action so the status can be updated
* Implement the new `reputation_override()` method so it can get a reputation override directly

This would have gone a lot smoother had I noticed that eliminating the status values from the search criteria does _not_ search for all statuses, it only searches for NEW status.  Once I got explicit about supplying _all_ the status values, the searches did more of what I expected them to.

The workflow test looks for a "candidate" recommendation on the server to apply tests to.  It chooses a NEW recommendation by preference, then a REJECTED recommendation, and lastly an ACCEPTED one.  The test makes sure it's left in the same state with which it entered the test.  Only if there are _no_ recommendations present on the server will it punt and say it can't do the test (as we have no way to _create_ recommendations on the fly).

avanbrunt-cb (approved):
LGTM

src/tests/uat/recommendation_uat.py
```diff
@@ -0,0 +1,423 @@
...
+        if verbose:
+            print("--- Accepting recommendation")
+        recommendation.accept()
+        result_workflow = validate_status_via_raw(recommendation, config_data) and result_workflow
+    elif initial_status == 'REJECTED':
+        if verbose:
+            print("--- Rejecting recommendation")
+        recommendation.reject()
+        result_workflow = validate_status_via_raw(recommendation, config_data) and result_workflow
+    return result_workflow
```
gyanitsa-cb, line 388: Is it possible by running this test to break another one that is depending on the selected recommendation if we don't reset it to its initial state?

abowersox-cb: There aren't any other tests involving recommendations. This is the only one.

src/tests/uat/recommendation_uat.py
```diff
@@ -0,0 +1,423 @@
...
+
+    Returns:
+        bool: True if comparison was successful, False if not.
+    """
+    result = compare(context, impact.event_count, raw_data, 'event_count')
+    result = compare(context, impact.impact_score, raw_data, 'impact_score') and result
+    result = compare(context, impact.impacted_devices, raw_data, 'impacted_devices') and result
+    result = compare(context, impact.org_adoption, raw_data, 'org_adoption') and result
+    result = compare(context, impact.update_time, raw_data, 'update_time') and result
+    return result
```
gyanitsa-cb, line 91: Can you print a message if 'result' is false before returning it? In this case, the message can be something like 'Comparing impact failed' or similar. Thus, in case of test failures, it would be easier to spot the exact failing step.

src/tests/uat/recommendation_uat.py
```diff
@@ -0,0 +1,423 @@
...
+        context (str): The context of the comparison, used for printing error messages.
+        application (RecommendationApplication): The Recommendation object to compare.
+        raw_data (dict): The dictionary containing what should be the same data.
+
+    Returns:
+        bool: True if comparison was successful, False if not.
+    """
+    result = compare(context, application.type, raw_data, 'type')
+    result = compare(context, application.value, raw_data, 'value') and result
+    return result
```
gyanitsa-cb, line 108: Can you print a message if 'result' is false before returning it?

src/tests/uat/recommendation_uat.py
```diff
@@ -0,0 +1,423 @@
...
+    result = compare(context, new_rule.override_type, raw_data, 'override_type') and result
+    result = compare(context, new_rule.path, raw_data, 'path') and result
+    result = compare(context, new_rule.sha256_hash, raw_data, 'sha256_hash') and result
+    result = compare(context, new_rule.signed_by, raw_data, 'signed_by') and result
+    if new_rule.application_:
+        result = compare_application(f"{context}.application", new_rule.application_,
+                                     raw_data['application']) and result
+    else:
+        result = check_not_present(context, raw_data, 'application') and result
+    return result
```
gyanitsa-cb, line 138: Can you print a message if 'result' is false before returning it?

src/tests/uat/recommendation_uat.py
```diff
@@ -0,0 +1,423 @@
...
+    Returns:
+        bool: True if comparison was successful, False if not.
+    """
+    result = compare(context, workflow.changed_by, raw_data, 'changed_by')
+    result = compare(context, workflow.create_time, raw_data, 'create_time') and result
+    result = compare(context, workflow.ref_id, raw_data, 'ref_id') and result
+    result = compare(context, workflow.status, raw_data, 'status') and result
+    result = compare(context, workflow.update_time, raw_data, 'update_time') and result
+    result = compare(context, workflow.comment, raw_data, 'comment') and result
+    return result
```
gyanitsa-cb, line 159: Can you print a message if 'result' is false before returning it?

src/tests/uat/recommendation_uat.py
```diff
@@ -0,0 +1,423 @@
...
+        result = check_not_present(context, raw_data, 'impact') and result
+    if recommendation.new_rule_:
+        result = compare_new_rule(f"{context}.new_rule", recommendation.new_rule_, raw_data['new_rule']) and result
+    else:
+        result = check_not_present(context, raw_data, 'new_rule') and result
+    if recommendation.workflow_:
+        result = compare_workflow(f"{context}.workflow", recommendation.workflow_, raw_data['workflow']) and result
+    else:
+        result = check_not_present(context, raw_data, 'workflow') and result
+    return result
```
gyanitsa-cb, line 189: Can you print a message if 'result' is false before returning it?

src/tests/uat/recommendation_uat.py
gyanitsa-cb: I mean of another team. For example, if we run this test on test03 during bug bash on the same org as other teams, can we break their tests?

abowersox-cb: Possible, but how likely is that to happen? There's certainly no way to "sync" between two running instances of the test.

src/tests/uat/recommendation_uat.py
abowersox-cb: The `compare()` function does that.

```
 if not result:
        print(f"field value {raw_field} did not match in {context} - object value {object_field}, "
              f"raw value {raw_data.get(raw_field, None)}")
```

src/tests/uat/recommendation_uat.py
abowersox-cb: The `compare()` function does that.

```
 if not result:
        print(f"field value {raw_field} did not match in {context} - object value {object_field}, "
              f"raw value {raw_data.get(raw_field, None)}")
```

src/tests/uat/recommendation_uat.py
abowersox-cb: The `compare()` function does that.

```
 if not result:
        print(f"field value {raw_field} did not match in {context} - object value {object_field}, "
              f"raw value {raw_data.get(raw_field, None)}")
```

src/tests/uat/recommendation_uat.py
abowersox-cb: The `compare()` function does that.

```
 if not result:
        print(f"field value {raw_field} did not match in {context} - object value {object_field}, "
              f"raw value {raw_data.get(raw_field, None)}")
```

src/tests/uat/recommendation_uat.py
abowersox-cb: The `compare()` function does that.

```
 if not result:
        print(f"field value {raw_field} did not match in {context} - object value {object_field}, "
              f"raw value {raw_data.get(raw_field, None)}")
```

src/tests/uat/recommendation_uat.py
gyanitsa-cb: I agree that it would probably be a rare situation. And I am thinking... even if we use a recommendation that a different simultaneously running test also attempts to use, resetting the state after the test will not do much to help. Cause they won't wait in order. I'm ok to leave it as is.

src/tests/uat/recommendation_uat.py
gyanitsa-cb: Thanks! I've somehow missed it.

gyanitsa-cb (approved):

PR #214, opened 2021-09-21
CBAPI-2383 - User guide - tutorial in RTD

abowersox-cb:
## Pull request checklist

https://jira.carbonblack.local/browse/CBAPI-2383

## Pull Request Description

Added Guide page for Recommendations API.

- [ ] Yes
- [x] No

## How Has This Been Tested?

Docs checked in RST formatter.

### DIFF
--- a/docs/recommendations.rst
+++ b/docs/recommendations.rst
@@ -1,0 +1,81 @@
+Recommendations
+===============
+
+Recommendations offer a quick shortcut for helping tune your policy configurations in an environment, by providing
+suggested reputation overrides which you may add to improve your policy. They can speed up the process of tuning your
+policy to an environment, rather than having to manually investigate endpoint activity and reconfigure the policy in
+response to those investigations.
+
+The Carbon Black Cloud SDK for Python offers assistance for dealing with Recommendations.
+
+Getting the List of Recommendations
+-----------------------------------
+
+By querying the ``Recommendation`` object, you can see which recommendations have already been generated for you by
+the Carbon Black Cloud.
+
+::
+
+    >>> from cbc_sdk import CBCloudAPI
+    >>> api = CBCloudAPI(profile='sample')
+    >>> from cbc_sdk.endpoint_standard import Recommendation
+    >>> query = api.select(Recommendation).set_statuses(['NEW', 'ACCEPTED', 'REJECTED']).sort_by('impact_score', 'DESC')
+    >>> recslist = list(query)
+    >>> for rec in recslist:
+    ...     print(f"{rec.recommendation_id} - {rec.workflow_.status} - {rec.impact_.impact_score}")
+    ...
+    5e6926d4-0c55-4757-a94d-e05883d5ee4c - ACCEPTED - 1.1710311
+    c9221b98-f64c-45dc-acb4-93caacd9dcee - ACCEPTED - 0.71904415
+    8afdcebc-12ca-4b63-8d8f-2d8055b652f2 - NEW - 0.6051892
+    100503cd-1897-425f-93b5-1ccba320438d - ACCEPTED - 0.2678737
+    b44586ea-4bba-4684-addb-d4934c10b3e2 - NEW - 0.1663951
+    3f89a837-034c-4b81-9f4c-f673a36ccb5c - NEW - 0.11177378
+    d9a9a122-8440-44d2-9fa4-84e77345155b - ACCEPTED - 0.08829709
+    7fcb6092-7adb-4069-861b-fed439d5f22a - REJECTED - 0.063399866
+    26ddb565-aff6-4b68-895c-fc286aa5f101 - ACCEPTED - 0.05499694
+    908e691e-5dbf-4028-bb52-a41cedada2f9 - ACCEPTED - 0.016786003
+
+**N.B.:** If you do not set status values on the recommendation query with ``set_statuses()``, the search defaults to
+looking for ``NEW`` recommendations *only.*
+
+Recommendations Workflow
+------------------------
+
+Individual recommendations in the ``NEW`` state may be accepted or rejected by calling their ``accept()`` or
+``reject()`` methods, respectively.
+
+::
+
+    # Continued from first example
+    >>> sublist = [rec for rec in recslist if rec.workflow_.status == 'NEW']
+    >>> sublist[0].accept('Comment for acceptance')
+    >>> print(sublist[0].workflow_.status)
+    ACCEPTED
+    >>> sublist[1].reject('Comment for rejection')
+    >>> print(sublist[1].workflow_.status)
+    REJECTED
+
+Individual recommendations in the ``ACCEPTED`` or ``REJECTED`` states may be reverted to the ``NEW`` state by calling
+their ``reset()`` method.
+
+::
+
+    # Continued from first example
+    >>> sublist = [rec for rec in recslist if rec.workflow_.status == 'REJECTED']
+    >>> sublist[0].reset()
+    >>> print(sublist[0].workflow_.status)
+    NEW
+
+Recommendations and Reputation Overrides
+----------------------------------------
+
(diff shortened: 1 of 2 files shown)


docs/recommendations.rst
```diff
@@ -0,0 +1,81 @@
...
+
+    >>> from cbc_sdk import CBCloudAPI
+    >>> api = CBCloudAPI(profile='sample')
+    >>> from cbc_sdk.endpoint_standard import Recommendation
+    >>> query = api.select(Recommendation).set_statuses(['NEW', 'ACCEPTED', 'REJECTED']).sort_by('impact_score', 'DESC')
+    >>> recslist = list(query)
+    >>> for rec in recslist:
+    ...     print(f"{rec.recommendation_id} - {rec.workflow_.status} - {rec.impact_.impact_score}")
+    ...
+    5e6926d4-0c55-4757-a94d-e05883d5ee4c - ACCEPTED - 1.1710311
```
mikethecanuck-cb, line 27: I don't know if this is possible, but for an API consumer it would be far superior to display the recommendation text than the recommendation ID. The ID is necessary for performing operations, but I can't make an effective decision which to leverage if I don't know what it's recommending.

Unfortunately because I can't find any recommendations in any of my test orgs I'm not sure what exactly to suggest here.

docs/recommendations.rst
```diff
@@ -0,0 +1,81 @@
...
+Recommendations Workflow
+------------------------
+
+Individual recommendations in the ``NEW`` state may be accepted or rejected by calling their ``accept()`` or
+``reject()`` methods, respectively.
+
+::
+
+    # Continued from first example
+    >>> sublist = [rec for rec in recslist if rec.workflow_.status == 'NEW']
```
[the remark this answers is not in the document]
mikethecanuck-cb: That's def some clever Python, but I definitely wouldn't have known what it's actually doing so I would've just closed my eyes and prayed that I never had to alter it

docs/recommendations.rst
```diff
@@ -0,0 +1,81 @@
...
+
+A recommendation in the ``ACCEPTED`` state will have a reputation override created for it.  You can retrieve that
+object with the ``reputation_override()`` method.
+
+::
+
+    # Continued from first example
+    >>> sublist = [rec for rec in recslist if rec.workflow_.status == 'ACCEPTED']
+    >>> reputation_override = sublist[0].reputation_override()
+    >>> print(reputation_override.id)
```
mikethecanuck-cb, line 80: Here too, printing the reputation override internal ID is a good start (if I just want to catalogue this for later), but it would be even better (especially for demonstration code like this, that helps me as a reader understand what the code's actually doing) to show the override's metadata - the list, type, description and value

docs/recommendations.rst
abowersox-cb: This is just an _example._ It is not intended to teach the reader _everything_ about dealing with recommendations, just show how the API accesses them.

docs/recommendations.rst
abowersox-cb: Again, this is just an _example._ Besides, the details of reputation overrides are beyond the scope of this guide, which deals with _recommendations._

avanbrunt-cb: I would disagree Reputation Overrides is a key detail of Recommendations. The guide doesn't need to be specifically targeted instead it should focus on the entire experience of the feature

avanbrunt-cb: You could also change the print to be `>>> print(reputation_override)` as I believe we have `__str__` functions in the base class which does a better job of showing more of the objects details

avanbrunt-cb: It's likely that people will copy directly from a guide to try something so having valuable content is key

docs/recommendations.rst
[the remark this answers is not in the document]
abowersox-cb: All the examples have been expanded so they don't depend on the first one.  (I executed the code that way in the first place to save time not having to run more queries, but it can be done another way.)

docs/recommendations.rst
abowersox-cb: > You could also change the print to be `>>> print(reputation_override)` as I believe we have `__str__` functions in the base class which does a better job of showing more of the objects details

Yeah, I saw that, but the output looked ugly, so I switched to just printing the ID.

avanbrunt-cb: If the output looked ugly then does that need fixed for this specific object? We can override the `__str__` method

abowersox-cb: No, I just think the way we output the objects is kind of ugly to put in example pages. But I see Reputation Override includes at least one on its page, so I'll change over to doing that if it'll let this page get past review.

emitreva-cb (approved):

docs/recommendations.rst
mikethecanuck-cb: I appreciate that examples can't spell out everything, but on the assumption that some users will copy our code and run it as-is - and to illustrate something more recognizable to the user (make the connection to the thing we're pulling out with this script - nearly any other attribute would be more relevant/illustrative than the random identifier value. If it'd help for me to suggest something, I'd love to help - but I don't know of an org where there are Recommendations *and* where I can generate an API key.

docs/recommendations.rst
mikethecanuck-cb: If I'm following the updated content correctly, looks like it's now "calling" print(reputation_override) and showing a pretty-printed columnular-ish display of all the properties of the outcome of accepting the recommendation - which is awesome. Thank you!

docs/recommendations.rst
abowersox-cb: I have revised this example to include also the filename and override list fields from the new rule.  This is the best I can do for you; it's _hard_ finding a combination of fields that will convey useful information, given how many of the Recommendations in our dev organization have "junk" values for some of those fields (like `None`). It took me four tries before I came up with _this_ combination, and, even then, I had to cut a _lot_ of the returned records because they had `None` in one of the columns, like the filename one.

avanbrunt-cb: What does it look like when you do `print(rec)`? Given the subclass structure of the object does it print all of the necessary fields? Does this class need a custom `__str__` method?

abowersox-cb: That is what we call a Darn Good Question. Fortunately, I have an answer:
[output_list.txt](https://github.com/carbonblack/carbon-black-cloud-sdk-python/files/7237964/output_list.txt)

As you can see, this output is both (a) truncated in the case of the subobjects, and (b) _way_ too long to include as sample output in the middle of a guide doc, even if I edited it down.

avanbrunt-cb: I would consider overriding the `__str__` method for Recommendation and implement something that does a better job of printing that. Similar to what we need to `Process.Tree` and `Process.Summary`

abowersox-cb: It's a little late in the game to go back and change the code, isn't it?

avanbrunt-cb: I'm not asking for something to change but rather to make an additive change to include a `__str__` function in the Recommendation class to better print that what is currently being exposed

mikethecanuck-cb: On the inline example front you nailed it @abowersox-cb - thank you for pawing through the data to show something that's more immediately recognizable as the kinds of outcomes the API allows.

I can see the value of such a `__str__` function, and would gladly support that if it could accompany this guidance.

abowersox-cb: Is there a way we can just declare what we have as Good Enough For Now, finish this ticket, and then create another one to deal with the `__str__` issue? Because putting that in seems to be going _way_ out of scope for this particular ticket.

avanbrunt-cb: Sure. As long as it is finished before the release

avanbrunt-cb (approved):

PR #221, opened 2021-10-04
Adding macOS's Keychain Access as an authentication handler

dganev-cb:
## Pull request checklist

Please check if your PR fulfils the following requirements:

https://jira.carbonblack.local/browse/CBAPI-2273

## Pull Request Description
Adding macOS's keychain as an alternative to our other authenticators.

- [x] No

## How Has This Been Tested?
Manually, using automated tests.


dganev-cb:
The thing that breaks the redhat build is the cryptography package. The latest versions require `rust-setuptools` and a rust compiler into the system. 

The `keyring` dependency requires the package. REF: https://github.com/pyca/cryptography/issues/5771

dganev-cb:
I've suggested to the author to split the dependencies into extras: https://github.com/jaraco/keyring/issues/532

setup.py
```diff
@@ -26,7 +26,8 @@
     'protobuf',
     'schema',
     'solrq',
-    'validators'
+    'validators',
+    'keyring'
```
avanbrunt-cb, line 30: Have you considered making this a platform specific dependency? `MacOS == Darwin` See https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#platform-specific-dependencies

dganev-cb: Yes, I completely forgot about this, thank you. I will add a comment for the future in case we want to support other platforms and tools such as Linux or other specific backends since that keyring supports a lot of credential managers.

bin/set-macos-keychain.py
```diff
@@ -0,0 +1,209 @@
...
+        file (str | Path): The cbc credential file.
+    """
+    parser = configparser.ConfigParser()
+    parser.read(file)
+    for section in parser.sections():
+        # reusing the parsing of the credentials because it is redundant to implement it twice
+        credentials = FileCredentialProvider(credential_file=file).get_credentials(section=section)
+        json_obj = {}
+        # replacing the CredentialValue's __str__ representation to get the default names.
+        for k, v in vars(credentials)["_values"].items():
```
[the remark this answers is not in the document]
dganev-cb: Yup, that would make more sense since this is a bit hacky.

bin/set-macos-keychain.py
```diff
@@ -0,0 +1,209 @@
...
+
+    Args:
+        args (Namespace): The parsed args from the ArgumentParser.
+    """
+    _args = vars(args)
+    if not _args["keychain_entry"]:
+        _args["keychain_entry"] = "CBC SDK API"
+    if not _args["keychain_account"]:
+        _args["keychain_account"] = "default"
+    keyring.set_password(_args["keychain_entry"], _args["keychain_account"], json.dumps(_args))
```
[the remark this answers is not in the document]
avanbrunt-cb: As you are dumping the entirety of `_args` into the value of the keychain entry this is the only case where you include the `keychain_account` and `keychain_entry` as properties/keys within the value of the entry. I would recommend not adding defaults values to the object and instead extract out the values into local variables like so

```
    _args = vars(args)
    keychain_entry = _args.pop("keychain_entry", "CBC SDK API")
    keychain_account = _args.pop("keychain_account", "default")
```

setup.py
avanbrunt-cb: You initial attempt was only a partial implementation. See this PR https://github.com/carbonblack/carbon-black-cloud-sdk-python/pull/225 on how it could be performed. However let's raise this in stand up tomorrow to discuss adding support for more than macOS as the keyring package supports other OS keyrings

Issue #217, opened 2021-09-27
[BUG] Bulk user deletion cannot be accomplished - grants are not deleted


dario-br:
I am unable to delete a user because grants are not deleted previously:

- OS: Windows 10
- Carbon Black Cloud Products: Platform (Users)
- Python Version: 3.6.5

**Describe the bug**
When I try to delete a list of users I receive a 409 error from the API indicating me that the grants should be deleted previously. Bulk deletion method for User class does not check previously for existing Grants.

**Steps to Reproduce**
```
cb = CBCloudAPI(profile=profile)
# Searches for all the users with the given email addresses
user_list = cb.select(User).email_addresses(email_list).all()
User.bulk_delete(user_list)
```

`cbc_sdk.errors.ClientError: Received error code 409 from API: {"success":false,"message":"The user's grant should be deleted first"}
`
**Expected behavior**
First, delete grants associated to each user and then proceed with the user deletion.

**Workaround** 
Code should do something like, User.bulk_delete:
```
[..]
        for user in users:
            grant = user.grant()
            if grant:
                grant.delete()
            user.delete()
```

abowersox-cb:
PR #233 fixes this issue and has been merged to develop, so it will be included in the next release.

PR #233, opened 2021-10-19
CBAPI-3122: Delete User Grant Before Deleting User

abowersox-cb:
## Pull request checklist

https://jira.carbonblack.local/browse/CBAPI-3122 (also GitHub #217)

## Pull Request Description

The delete() method on User was failing if the user had an outstanding access grant.  We modify the delete() to check for an access grant on the user, and, if one exists, delete the grant before deleting the user.

- [ ] Yes
- [x] No

## How Has This Been Tested?

New unit tests have been written for the User.delete(), covering both the cases where the user has and does not have a grant. The test for User.bulk_delete() has also been updated.

The script 
[delete_user.py.txt](https://github.com/carbonblack/carbon-black-cloud-sdk-python/files/7376862/delete_user.py.txt) is a sample script that deletes a user from the command line given the user's E-mail address. The text log
[before-and-after.txt](https://github.com/carbonblack/carbon-black-cloud-sdk-python/files/7376863/before-and-after.txt) shows the output from that script both before and after the fix to User.delete() was applied.

### DIFF
--- a/bin/set-macos-keychain.py
+++ b/bin/set-macos-keychain.py
@@ -11,17 +11,15 @@
 # * WARRANTIES OR CONDITIONS OF MERCHANTABILITY, SATISFACTORY QUALITY,
 # * NON-INFRINGEMENT AND FITNESS FOR A PARTICULAR PURPOSE.
 
-__doc__ = """
+"""
 Helper script for adding an entry into macos's keychain.
 
 Use the script by simply invoking it without any system arguments to
 import your configuration from a file or manually by inputs.
 
-Invoke it with system arguments to create an entry without having to type the credentials
-manually.
+Invoke it with system arguments to create an entry without having to type the credentials manually.
 
 Examples:
-
     Using system arguments:
     $ ./set-macos-keychain.py -e CBC_SDK -a default -t <TOKEN> -k <ORG_KEY> -u <URL>
 
--- a/src/tests/unit/platform/test_users.py
+++ b/src/tests/unit/platform/test_users.py
@@ -13,7 +13,7 @@ from tests.unit.fixtures.platform.mock_users import (GET_USERS_RESP, GET_USERS_A
                                                      EXPECT_USER_ADD, EXPECT_USER_ADD_SMALL, EXPECT_USER_ADD_V1,
                                                      EXPECT_USER_ADD_V2, EXPECT_USER_ADD_BULK1, EXPECT_USER_ADD_BULK2,
                                                      USER_ADD_SUCCESS_RESP, USER_ADD_FAILURE_RESP)
-from tests.unit.fixtures.platform.mock_grants import (DETAILS_GRANT1, EXPECT_CHANGE_ROLE_GRANT1,
+from tests.unit.fixtures.platform.mock_grants import (DETAILS_GRANT1, EXPECT_CHANGE_ROLE_GRANT1, EXPECT_DELETE_GRANT1,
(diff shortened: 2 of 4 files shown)


bin/set-macos-keychain.py
```diff
@@ -11,17 +11,15 @@
 # * WARRANTIES OR CONDITIONS OF MERCHANTABILITY, SATISFACTORY QUALITY,
 # * NON-INFRINGEMENT AND FITNESS FOR A PARTICULAR PURPOSE.
 
-__doc__ = """
```
abowersox-cb, line 14: Changes in this file are only because it showed flake8 errors when I ran flake8 across the entire project.

src/tests/unit/platform/test_users.py
```diff
@@ -94,20 +94,6 @@ def check_put(url, body, **kwargs):
     assert user.email == 'jsheridan@zhadum.net'
 
 
-def test_get_and_delete_user(cbcsdk_mock):
```
abowersox-cb, line 97: This test was removed because it is replaced by two tests, below.

emitreva-cb (approved):

src/tests/unit/platform/test_users.py
```diff
@@ -269,6 +255,27 @@ def test_get_grant(cbcsdk_mock):
     assert grant.roles == ["psc:role::SECOPS_ROLE_MANAGER", "psc:role:test:APP_SERVICE_ROLE"]
 
 
+def test_delete_user_without_grant(cbcsdk_mock):
```
dganev-cb, line 258: I think those tests will always pass since there is no assertion or anything that checks if the user is actually deleted.

dganev-cb (changes requested):

src/tests/unit/platform/test_users.py
abowersox-cb: _Any_ exception will be a "fail" on the test. There shouldn't be a need to check for a specific one.

src/tests/unit/platform/test_users.py
```diff
@@ -269,6 +255,27 @@ def test_get_grant(cbcsdk_mock):
...
+    """Tests deleting a user that does not have a grant."""
+    cbcsdk_mock.mock_request('GET', '/appservices/v6/orgs/test/users', GET_USERS_RESP)
+    cbcsdk_mock.mock_request('POST', '/access/v2/grants/_fetch', wrap_grant(None))
+    cbcsdk_mock.mock_request('DELETE', '/appservices/v6/orgs/test/users/3911', None)
+    api = cbcsdk_mock.api
+    user = api.select(User).user_ids([3911]).one()
```
avanbrunt-cb, line 264: This should use our better practice like so
```
user = api.select(User, 3911)
```

src/tests/unit/platform/test_users.py
abowersox-cb: This has been done. Take the new delete() code out now and the two tests will both fail with the assert message, "Grant should have been checked for prior to deleting user".

src/tests/unit/platform/test_users.py
abowersox-cb: Actually, I'm wrong, the code has a contingency for that by calling _refresh(), which knows how to fake it. I'll change this.

avanbrunt-cb: Perfect!

avanbrunt-cb (approved):
LGTM. Thanks for the extra verification

PR #234, opened 2021-10-21
Create dynamic referencing of classes

dganev-cb:
## Pull request checklist

Please check if your PR fulfils the following requirements:

3121

## Pull Request Description
I think it will be a good addition to our ORM-like SDK and since its widely used in the Python community in frameworks such as Django's ORM, Flask's Blueprints and etc. 

And when I first interacted with the SDK and saw how it works I thought it would have that functionality, simply sharing this to give you my perspective of how I saw the SDK at first. Definitely needs more research from our clients since they are not heavy python users.

Referer to the other argumentation in the ticket: 3121 

- [x] No

## How Has This Been Tested?
Currently the tests has been partially done for the Alerts only. The testing must cover all of our queryable models.

### DIFF
--- a/docs/alerts.rst
+++ b/docs/alerts.rst
@@ -237,4 +237,4 @@ Those settings shown in the screenshot can be replicated with the following code
 Advanced usage of alerts
 ------------------------
 
-If you want near-real-time streaming of an alerts we advice you to referer our `Data Forwarded <https://developer.carbonblack.com/reference/carbon-black-cloud/platform/latest/data-forwarder-config-api/>`_.
+If you want near-real-time streaming of alerts we advise you to refer to our `Data Forwarder <https://developer.carbonblack.com/reference/carbon-black-cloud/platform/latest/data-forwarder-api/>`_.
--- a/docs/authentication.rst
+++ b/docs/authentication.rst
@@ -34,9 +34,9 @@ Store the Key with profile name, and reference the profile name when creating CB
 
   # execute commands with Live Response API
   >>> for device in important_devs:
-  ...      lr_session = platform_api.live_response.request_session(device.id)
-  ...      lr_session.create_process(r'cmd.exe /c "ping.exe 192.168.1.1"'))
-  ...      lr_session.close()
+  ...     lr_session = platform_api.live_response.request_session(device.id)
+  ...     lr_session.create_process(r'cmd.exe /c "ping.exe 192.168.1.1"')
+  ...     lr_session.close()
 
 For more examples on Live Response, check :doc:`live-response`
 
@@ -92,8 +92,8 @@ Authentication Methods
   via keyword parameters. This method should be used with caution, taking care to not
   share your API credentials when managing code with source control.
 
-    >>> cbc_api = CBCloudAPI(url='defense.conferdeploy.net', token=ABCD/1234,
-    ...                         org_key='ABCDEFGH')
+    >>> cbc_api = CBCloudAPI(url='https://defense.conferdeploy.net', token='ABCD/1234',
+    ...     org_key='ABCDEFGH')
 
 **Not Recommended**:
 
--- a/docs/concepts.rst
+++ b/docs/concepts.rst
@@ -53,11 +53,10 @@ to USB devices is through the Endpoint Standard package ``from cbc_sdk.endpoint_
   >>> api = CBCloudAPI(profile='endpoint_standard')
   >>> usb_devices = api.select(USBDevice).set_statuses(['APPROVED'])
   >>> for usb in usb_devices:
-  ...   print(f'''
-              USB Device ID: {usb.id}
-              USB Device: {usb.vendor_name} {usb.product_name}
-
-              ''')
+  ...     print(f'''
+  ...         USB Device ID: {usb.id}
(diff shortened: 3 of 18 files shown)


dganev-cb:
I've added a testing example of how it might be tested. @abowersox-cb can you check it out and tell me what do you think about it. I've decided that `__qualname__` would be a better alternative than just `__name__` since we have nested classes.

dganev-cb:
> I've added a testing example of how it might be tested. @abowersox-cb can you check it out and tell me what do you think about it. I've decided that `__qualname__` would be a better alternative than just `__name__` since we have nested classes.

Actually now thinking about it I should probably only test the `select_class_instance` function since the rest of the logic remains the same in the `select` gonna update the unit tests.

abowersox-cb:
Can you fix the flake8 errors? I want to make sure the new unit tests run before I approve this, because I'm still not sure how you're dealing with the module-loading issue.

dganev-cb:
> Can you fix the flake8 errors? I want to make sure the new unit tests run before I approve this, because I'm still not sure how you're dealing with the module-loading issue.

I will do that asap, I am still figuring out why those modules are not loading  `SensorKit`, `ComputeResource`.

dganev-cb:
@abowersox-cb @avanbrunt-cb It turns out that, the `SensorKit, ComputeResource` are the only 2 modules that are not loading. I've looked for other solutions and currently, I have 2.

1. We import them in the `__init__.py` level and with that we can have it loaded.
2. Go with an entirely different solution by using a `dict` with keys as the classes and values with `import_module` function. 

I don't like either of those and I've asked around to see if I can find some better solutions. I will put this into a stall for now.

dganev-cb:
@abowersox-cb @avanbrunt-cb I've added a list variable with `import_module` function just to make sure that everytime we call `select_class_instance` all of our modules are imported. This way we won't need to address an addition or a change whenever we create new models. 

I've also changed the name of the Event in the Endpoint Standard to `EndPointStandardEvent` since that I am not sure if that was the model that is going to be deprecated I will leave this at this stage. Maybe a more appropriate name or an eventual usage of a `@deprecated` decorator can be used.

dganev-cb:
Since I found new information over this feature I am updating it.

We can use the `pkgutil.walk_packages` to iterate over all the modules located in the cbc_sdk and load them. 
(https://docs.python.org/3/library/pkgutil.html#pkgutil.walk_packages) 

The only thing that is not tested is the `Platform.Event` because it is going to be deprecated, it clashes with the `endpoint_standard.Event`. 

Tell me what do you think @abowersox-cb @emitreva-cb @avanbrunt-cb

avanbrunt-cb (commented):
We should not merge this until we remove the event code which should be done by the end of January given that functionality will no longer be available in the CBC

requirements.txt
```diff
@@ -12,6 +12,7 @@ solrq
 validators
 keyring;platform_system=='Darwin'
 boto3
+deprecation
```
avanbrunt-cb, line 15: I don't see where this is being used. We need to be careful adding too many new packages.

src/tests/unit/base/test_queries.py
```diff
@@ -70,6 +73,77 @@ def _get_results(url, **kwargs):
...
+        ("Report", "ReportQuery"),
+        ("Watchlist", "WatchlistQuery"),
+
+        # Platform
+        ("BaseAlert", "BaseAlertSearchQuery"),
+        ("CBAnalyticsAlert", "CBAnalyticsAlertSearchQuery"),
+        ("DeviceControlAlert", "DeviceControlAlertSearchQuery"),
+        ("WatchlistAlert", "WatchlistAlertSearchQuery"),
+        ("Device", "DeviceSearchQuery"),
+        ("Event", "EventQuery"),
```
avanbrunt-cb, line 123: Using the following code does not reference the correct Event model
```
cb.select('Event', '7DESJ9GN-0040bdb5-000009e0-00000000-1d82fe3b49a5b32')
```

instead it matches to the following
```
<cbc_sdk.endpoint_standard.base.Event: id 7DESJ9GN-0040bdb5-000009e0-00000000-1d82fe3b49a5b32> @ https://defense.conferdeploy.net
```

Was expecting this instead
```
<cbc_sdk.platform.events.Event: id 7DESJ9GN-0040bdb5-000009e0-00000000-1d82fe3b49a5b32> @ https://defense.conferdeploy.net
```

avanbrunt-cb (changes requested):

dganev-cb:
@avanbrunt-cb I cannot replicate that, and this shouldn't be possible since the `Event` that you are referring to doesn't extend the `NewBaseModel`.

avanbrunt-cb (approved):
I've modified the test file as the `_test` name caused the tests to be ignored. I've also enhanced the event test to include the other approach of specifying the process_guid in order to ensure both mechanisms import the platform models. Lastly, the deprecation package was still included in the requirements.txt file which was not being utilized.

I am comfortable with the changes now after further testing and ensuring I had the latest changes

abowersox-cb (approved):
Good work cleaning this up, @avanbrunt-cb.

PR #239, opened 2021-11-11
CBAPI-3169: CBC SDK Live Query Ease of Use Improvements

abowersox-cb:
## Pull request checklist

https://jira.carbonblack.local/browse/CBAPI-3169

## Pull Request Description

Added several functions to LiveQuery support to make it easier to get runs for a template and results, device summaries, or facets from a run.  Also added asynchronous query support to all LiveQuery queries (including the LiveQuery itself).

Also checked over all docstrings in the LiveQuery code and updated the unit tests that used older mocking to use the CBCSDKMock.

- [ ] Yes
- [x] No

## How Has This Been Tested?

Unit tests added that cover all new code added to LiveQuery.  Coverage on complete source file is at 94%.

### DIFF
--- a/src/cbc_sdk/audit_remediation/base.py
+++ b/src/cbc_sdk/audit_remediation/base.py
@@ -31,7 +31,8 @@ MAX_RESULTS_LIMIT = 10000
 
 
 class Run(NewBaseModel):
-    """Represents an Audit and Remediation run.
+    """
+    Represents an Audit and Remediation run.
 
     Example:
     >>> run = cb.select(Run, run_id)
@@ -46,7 +47,17 @@ class Run(NewBaseModel):
     _is_deleted = False
 
     def __init__(self, cb, model_unique_id=None, initial_data=None):
-        """Initialize a Run object with initial_data."""
+        """
+        Initialize a Run object with initial_data.
+
+        Required Permissions:
+            livequery.manage(READ)
+
+        Arguments:
+            cb (BaseAPI): Reference to API object used to communicate with the server.
+            model_unique_id (str): ID of the query run represented.
+            initial_data (dict): Initial data used to populate the query run.
+        """
         if initial_data is not None:
             item = initial_data
         elif model_unique_id is not None:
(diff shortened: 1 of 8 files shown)


src/cbc_sdk/audit_remediation/base.py
```diff
@@ -588,8 +633,36 @@ def submit(self):
...
+        Executed in the background to run an asynchronous query. Must be implemented in any inheriting classes.
 
-class RunHistoryQuery(BaseQuery, QueryBuilderSupportMixin, IterableQueryMixin, CriteriaBuilderSupportMixin):
+        Args:
+            context (object): The context returned by _init_async_query. May be None.
+
+        Returns:
+            Any: Result of the async query, which is then returned by the future.
+        """
+        while context.status == 'ACTIVE':
```
dganev-cb, line 868: Wouldn't those states `ACTIVE, COMPLETE etc..` be better in an `enum`?

emitreva-cb (approved):

src/cbc_sdk/audit_remediation/base.py
abowersox-cb: To turn them into an enum now would break existing code, which we prefer not to do unless there's no way around it.

src/cbc_sdk/audit_remediation/base.py
```diff
@@ -588,11 +845,50 @@ def submit(self):
...
+    def _init_async_query(self):
+        """
+        Initialize an async query and return a context for running in the background. Optional.
+
+        Returns:
+            Run: Context for running in the background.
+        """
+        return self.submit()
+
+    def _run_async_query(self, context):
```
avanbrunt-cb, line 857: I'm not sure this makes sense given the time for a run to complete.

avanbrunt-cb (approved):

src/cbc_sdk/audit_remediation/base.py
abowersox-cb: Hmmm...not sure what I can say to that. The functionality is there, and it works in unit tests.

dganev-cb (approved):

PR #248, opened 2021-12-09
CBAPI-3328: SDK: Export Live Query Results - Line-by-Line Iteration

abowersox-cb:
## Pull request checklist

https://jira.carbonblack.local/browse/CBAPI-3328

## Pull Request Description

Added additional method to `ResultQuery`, `export_csv_as_lines()`, that returns the exported CSV as an iterator of lines.  By request of @avanbrunt-cb.

- [ ] Yes
- [x] No

## How Has This Been Tested?

New test script has been written that proves the download works against a live environment.  Test script (with internal data redacted) follows.
```
#!/usr/bin/env python

from cbc_sdk.audit_remediation import Run

api = CBCloudAPI(profile='[REDACTED]')
run = api.select(Run, '[REDACTED]')
for line in run.query_results().export_csv_as_lines():
    print(line)
```
Additional unit tests have been written for the new functionality.  All new code is covered.

### DIFF
--- a/src/cbc_sdk/audit_remediation/base.py
+++ b/src/cbc_sdk/audit_remediation/base.py
@@ -1385,13 +1385,13 @@ class ResultQuery(BaseQuery, QueryBuilderSupportMixin, IterableQueryMixin, Crite
             livequery.manage(READ)
 
         Returns:
-            iterator: An iterator that can be used to get each line of CSV text in turn as a string.
+            iterable: An iterable that can be used to get each line of CSV text in turn as a string.
         """
         if self._run_id is None:
             raise ApiError("Can't retrieve results without a run ID")
         url = self._doc_class.urlobject.format(self._cb.credentials.org_key, self._run_id) + '?format=csv'
         request = self._build_request(0, -1)
-        return self._cb.post_and_get_lines(url, request, headers={'Accept': 'text/csv'})
+        yield from self._cb.post_and_get_lines(url, request, headers={'Accept': 'text/csv'})
 
     def export_zipped_csv(self, filename):
         """
--- a/src/cbc_sdk/audit_remediation/base.py
+++ b/src/cbc_sdk/audit_remediation/base.py
@@ -1377,6 +1377,22 @@ class ResultQuery(BaseQuery, QueryBuilderSupportMixin, IterableQueryMixin, Crite
         with io.open(filename, 'wb') as file:
             self.export_csv_as_stream(file)
 
+    def export_csv_as_lines(self):
+        """
+        Export the results from the run as CSV, returning the CSV data as iterated lines.
+
+        Required Permissions:
+            livequery.manage(READ)
+
+        Returns:
+            iterator: An iterator that can be used to get each line of CSV text in turn as a string.
+        """
+        if self._run_id is None:
+            raise ApiError("Can't retrieve results without a run ID")
+        url = self._doc_class.urlobject.format(self._cb.credentials.org_key, self._run_id) + '?format=csv'
+        request = self._build_request(0, -1)
+        return self._cb.post_and_get_lines(url, request, headers={'Accept': 'text/csv'})
+
     def export_zipped_csv(self, filename):
         """
         Export the results from the run as a zipped CSV, writing the zip data to the named file.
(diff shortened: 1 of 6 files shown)


emitreva-cb (approved):

src/tests/unit/audit_remediation/test_audit_remediation_base.py
```diff
@@ -309,6 +309,17 @@ def test_result_query_export_file(cbcsdk_mock):
...
 
 
+def test_result_query_export_lines(cbcsdk_mock):
+    """Tests exporting the results of a query as a list of lines."""
+    input = "AAA\r\nBBB\r\nCCC"
+    cbcsdk_mock.mock_request("POST_LINES", "/livequery/v1/orgs/test/runs/run_id/results/_search?format=csv",
+                             CBCSDKMock.StubResponse(input, 200, input, False))
+    api = cbcsdk_mock.api
+    result_query = api.select(Result).run_id("run_id")
+    output = result_query.export_csv_as_lines()
```
avanbrunt-cb, line 319: I thought this was suppose to return an iterator not a list?

src/tests/unit/fixtures/stubresponse.py
```diff
@@ -49,6 +49,10 @@ def iter_content(self, buffer_size=-1):
             return_list.append(part)
         return return_list
 
+    def iter_lines(self, **kwargs):
+        """Create iterated lines from the 'content' parameter."""
+        return self.content.splitlines()
```
avanbrunt-cb, line 54: This should yield the results individually to match the expected functionality

src/tests/unit/audit_remediation/test_audit_remediation_base.py
abowersox-cb: Perhaps I misspoke. It returns, not an _iterator,_ but an _iterable._  I can fix the docstrings to reflect that.

src/tests/unit/fixtures/stubresponse.py
abowersox-cb: It shouldn't make one iota of difference functionally, but I will do so.

src/tests/unit/audit_remediation/test_audit_remediation_base.py
avanbrunt-cb: Have you run this against the CBC APIs? I'm curious if the mocking matches the behavior of a live run?

avanbrunt-cb: To verify my understanding, this won't read the entire file into memory instead it will process line by line as requested. If we load it into a list that would but it could be processed in a for loop to prevent that.

abowersox-cb: Oof, you're right.  I need to encapsulate the output in a list to match the equivalent test that's made at the low level. Doing so.

And yes, `export_csv_as_lines()` _has_ been run against a live CBC server. The test script I used is up in the PR description.

src/cbc_sdk/audit_remediation/base.py
```diff
@@ -1377,6 +1377,22 @@ def export_csv_as_file(self, filename):
...
+            livequery.manage(READ)
+
+        Returns:
+            iterator: An iterator that can be used to get each line of CSV text in turn as a string.
+        """
+        if self._run_id is None:
+            raise ApiError("Can't retrieve results without a run ID")
+        url = self._doc_class.urlobject.format(self._cb.credentials.org_key, self._run_id) + '?format=csv'
+        request = self._build_request(0, -1)
+        return self._cb.post_and_get_lines(url, request, headers={'Accept': 'text/csv'})
```
avanbrunt-cb, line 1394: This needs to be a yield to pass through the generator instead of converting it to a list

abowersox-cb: OK, thanks for the clarification on how that works.

avanbrunt-cb (approved):

src/tests/unit/audit_remediation/test_audit_remediation_base.py
```diff
@@ -334,6 +345,8 @@ def test_result_query_exports_fail_without_run_id(cb):
         result_query.export_csv_as_string()
     with pytest.raises(ApiError):
         result_query.export_csv_as_file('blort.txt')
+    with pytest.raises(ApiError):
+        list(result_query.export_csv_as_lines())
```
abowersox-cb, line 349: You didn't have to tell me what I needed to do here to make this work, @avanbrunt-cb.  I figured it out on my own. :)

PR #252, opened 2021-12-20
Add `boto3` to `setup.py`

dganev-cb:
## Pull request checklist
Add `boto3` to `setup.py`

CBAPI-2274 

- [x] No

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,8 @@ install_requires = [
     'schema',
     'solrq',
     'validators',
-    "keyring;platform_system=='Darwin'"
+    "keyring;platform_system=='Darwin'",
+    'boto3'
 ]
 
 tests_requires = [


emitreva-cb (approved):

dganev-cb:
Context (which I forgot to add):  Forgot to add the `boto3` requirement to our setup.py file.

avanbrunt-cb (approved):
As we add more possible credential handlers, I am curious if we want to continue to expand our imports which would increase our overall package size. Would be worth investigating optional flags or other mechanisms that could be used to opt in these extra packages if they want to utilize the other credential providers

dganev-cb:
> As we add more possible credential handlers, I am curious if we want to continue to expand our imports which would increase our overall package size. Would be worth investigating optional flags or other mechanisms that could be used to opt in these extra packages if they want to utilize the other credential providers

I believe something like extras: `pip install carbon-black-cloud-python-sdk[aws]` can be used for this, I will research it.

PR #254, opened 2021-12-24
Live query guide

kebringer-cb:
## Pull request checklist

- [X] Docs have been reviewed and added / updated if needed (for bug fixes / features)

## Pull Request Description
Live query guide for the SDK that shows how to do the same steps as the Quick Start on the Dev Network.

- [ ] Yes
- [X ] No

## How Has This Been Tested?
Generated docs locally. Ran the snippets top to bottom.

### DIFF
--- a/docs/live-query.rst
+++ b/docs/live-query.rst
@@ -9,10 +9,9 @@ using Live Query via the Python SDK.
 
 More information about the Audit and Remediation product which uses Live Query is available in the
 `Carbon Black Cloud user guide
-<https://docs.vmware.com/en/VMware-Carbon-Black-Cloud/services/carbon-black-cloud-user-guide/GUID-129D4F84-1BF0-49F3-BF95-83002FD63217.html>`_
+<https://docs.vmware.com/en/VMware-Carbon-Black-Cloud/services/carbon-black-cloud-user-guide/GUID-129D4F84-1BF0-49F3-BF95-83002FD63217.html/>`_
 
-More information about Live Query APIs is available on the `Developer Network
-<https://developer.carbonblack.com/reference/carbon-black-cloud/cb-liveops/>`'.
+More information about Live Query APIs is available on the `Developer Network <https://developer.carbonblack.com/reference/carbon-black-cloud/cb-liveops/>`_.
 
 Overview
 --------
--- a/docs/live-query.rst
+++ b/docs/live-query.rst
@@ -8,22 +8,23 @@ powered by https://osquery.io, an open source project that uses an SQLite interf
 using Live Query via the Python SDK.
 
 More information about the Audit and Remediation product which uses Live Query is available in the
-Carbon Black Cloud user guide:
-https://docs.vmware.com/en/VMware-Carbon-Black-Cloud/services/carbon-black-cloud-user-guide/GUID-129D4F84-1BF0-49F3-BF95-83002FD63217.html
+`Carbon Black Cloud user guide
+<https://docs.vmware.com/en/VMware-Carbon-Black-Cloud/services/carbon-black-cloud-user-guide/GUID-129D4F84-1BF0-49F3-BF95-83002FD63217.html>`_
 
-More information about Live Query APIs is available on the Developer Network.
-https://developer.carbonblack.com/reference/carbon-black-cloud/cb-liveops/
+More information about Live Query APIs is available on the `Developer Network
+<https://developer.carbonblack.com/reference/carbon-black-cloud/cb-liveops/>`'.
 
 Overview
 --------
 This guide shows how to find specific files on a system. This is the same scenario as the Quick Start Guide for the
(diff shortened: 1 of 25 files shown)


docs/live-query.rst
```diff
@@ -0,0 +1,100 @@
...
+-----------------
+Set up the query for the file you are looking for.  Then create a query object, execute it, and get the id of the run.
+
+    >>> query_string = "SELECT filename, path FROM file WHERE path = 'C:\\Windows\\Temp\\dbutil_2_3.sys\\' OR path LIKE 'C:\\Users\\%\\AppData\\Local\\Temp\\dbutil_2_3.sys';"
+    >>> query_object = api.select(Run).where(sql=query_string)
+    >>> run = query_object.submit()
+    >>> print(run.id)
+    vsc2be500dcuhc1q5bhvq7kdwoqh367i
+
+Get the Query Run to check status
```
avanbrunt-cb, line 50: For my context what is the intention  behind this section as you already have the run from the previous section as that is returned from the `submit()`. Are we needing  to refresh the object to  get the update status? We could use `run.refresh()`

docs/live-query.rst
```diff
@@ -0,0 +1,100 @@
...
+not_started_count: 40
+status: ACTIVE
+    total_results: 0
+
+Get the results
+---------------
+Partial results can be reviewed while the query is running.  This snippet gets the results and prints the device
+information for each.
+
+    >>> result_query = api.select(Result).run_id(run.id)
```
avanbrunt-cb, line 73: We also added a helper function on the run object if they don't want to construct this themselves `query_results()`. The helper function could be mentioned in a note as an alternative as I do like showing this option as the primary way too

docs/live-query.rst
```diff
@@ -0,0 +1,100 @@
...
+
+This snippet shows writing the results to a zipped csv file.
+
+   >>> api.select(Result).run_id(run.id).export_zipped_csv("/Users/myname/mydir/livequeryresults.zip")
+
+Clean up
+---------
+Since this is a tutorial we'll clean up when we're done by first stopping the query and then deleting it.
+
+    >>> run.stop()
```
avanbrunt-cb, line 95: I'd like us to add context that the stop will not prevent any devices currently running the query as it will only receive the stop at next check in so only the ones which have not started will not complete.

Something I am not 100% sure on but would make sense is that the results for that run would be deleted as well. If this is true we should mention this as well

lhilmes-cb (approved):
Can't validate technical information, but looks good to me

docs/live-query.rst
kebringer-cb: Changed to run.refresh()

docs/live-query.rst
kebringer-cb: added

docs/live-query.rst
kebringer-cb: done

docs/live-query.rst
```diff
@@ -0,0 +1,176 @@
...
+More information about the Audit and Remediation product which uses Live Query is available in the
+`Carbon Black Cloud user guide
+<https://docs.vmware.com/en/VMware-Carbon-Black-Cloud/services/carbon-black-cloud-user-guide/GUID-129D4F84-1BF0-49F3-BF95-83002FD63217.html/>`_
+
+More information about Live Query APIs is available on the `Developer Network <https://developer.carbonblack.com/reference/carbon-black-cloud/cb-liveops/>`_.
+
+Overview
+--------
+This guide shows how to find specific files on a system. This is the same scenario as the Quick Start Guide for the
+APIs on the `Developer Network <https://developer.carbonblack.com/reference/carbon-black-cloud/cb-liveops/latest/livequery-api/>`_
```
hkaragitliev-cb, line 19: You can point the link directly to the Quick Start section: https://developer.carbonblack.com/reference/carbon-black-cloud/cb-liveops/latest/livequery-api/#quick-start

And a note: I have added two more quick start guides for Live Query that will be available with the DIfferential Analysis API. 
1. Find Installed Firefox add-ons
2. Automatically Recurring Query

docs/live-query.rst
```diff
@@ -0,0 +1,176 @@
...
+#. Start the Query Run
+#. Look at the results
+#. Write the results to a file
+#. Clean up since this is a tutorial
+#. Get the run information for scheduled queries (templates)
+
+Setting up
+----------
+
+The code snippets assume that the python environment has been set up with the necessary imports and credentials.
```
hkaragitliev-cb, line 33: Maybe link the authentication section from here.

docs/live-query.rst
kebringer-cb: thanks.  I'll make a note to update this guide when Diff Analysis is added.

abowersox-cb (approved):
This looks like just what we need.

avanbrunt-cb (approved):

PR #256, opened 2022-01-10
CBAPI-3296: Live Query Async with Jobs Service

abowersox-cb:
## Pull request checklist

https://jira.carbonblack.local/browse/CBAPI-3296

## Pull Request Description

Added support for retrieving results for a Live Query via an asynchronous job.  Includes adding support for the Job Service API to the Platform part of the SDK. Job results may be retrieved as stream output, file download, one big string, or an iterated list of lines, as with existing LiveQuery results support.

- [ ] Yes
- [x] No

## How Has This Been Tested?

Job service has been run against PROD02 successfully. Unit tests have been added for all new code.  Coverage of new Jobs API is at 100%. No net loss of coverage in other portions of the SDK.  Test script run against PROD02 is as follows:
```
#!/usr/bin/env python

from cbc_sdk.audit_remediation import Run

api = CBCloudAPI(profile='prod02-super')
run = api.select(Run, 'uerrofnzcdgychxi9icquislzjhclswo')
job = run.query_results().async_export()
print(f"New job created with ID {job.id}")
future = job.await_completion()
for s in future.result().get_output_as_lines():
    print(s)
```

### DIFF
--- a/src/cbc_sdk/audit_remediation/base.py
+++ b/src/cbc_sdk/audit_remediation/base.py
@@ -1413,6 +1413,8 @@ class ResultQuery(BaseQuery, QueryBuilderSupportMixin, IterableQueryMixin, Crite
         """
         Create an asynchronous job that exports the results from the run.
 
+        This is recommended if you are expecting a very large result set.
+
         Required Permissions:
             livequery.manage(READ), jobs.status(READ)
 
--- a/src/cbc_sdk/audit_remediation/base.py
+++ b/src/cbc_sdk/audit_remediation/base.py
@@ -17,6 +17,7 @@ from __future__ import absolute_import
 from cbc_sdk.base import (UnrefreshableModel, NewBaseModel, QueryBuilder,
                           QueryBuilderSupportMixin, IterableQueryMixin, BaseQuery,
                           CriteriaBuilderSupportMixin, AsyncQueryMixin)
+from cbc_sdk.platform import Job
 from cbc_sdk.errors import ApiError, ServerError, TimeoutError, OperationCancelled
 import io
 import logging
@@ -1407,6 +1408,30 @@ class ResultQuery(BaseQuery, QueryBuilderSupportMixin, IterableQueryMixin, Crite
             raise ApiError("Can't retrieve results without a run ID")
         with io.open(filename, 'wb') as file:
             self.export_csv_as_stream(file, True)
+
+    def async_export(self):
+        """
+        Create an asynchronous job that exports the results from the run.
+
+        Required Permissions:
+            livequery.manage(READ), jobs.status(READ)
+
+        Returns:
+            Job: The Job object that represents the asynchronous job.
+        """
+        if self._run_id is None:
+            raise ApiError("Can't retrieve results without a run ID")
+        url = self._doc_class.urlobject.format(self._cb.credentials.org_key, self._run_id) + '?format=csv&async=true'
+        request = self._build_request(0, -1)
+        response = self._cb.post_object(url, request)
+        ref_url = response.get('ref_url', None)
+        try:
+            job_id = int(ref_url.rsplit('/', 1)[1]) if ref_url else -1
+        except ValueError:
+            job_id = -1
+        if job_id < 0:
+            raise ApiError(f"server sent back invalid job reference URL {ref_url}")
+        return Job(self._cb, job_id)
 
 
 class FacetQuery(BaseQuery, QueryBuilderSupportMixin, IterableQueryMixin, CriteriaBuilderSupportMixin, AsyncQueryMixin):
--- a/src/cbc_sdk/audit_remediation/base.py
+++ b/src/cbc_sdk/audit_remediation/base.py
@@ -1424,7 +1424,7 @@ class ResultQuery(BaseQuery, QueryBuilderSupportMixin, IterableQueryMixin, Crite
         url = self._doc_class.urlobject.format(self._cb.credentials.org_key, self._run_id) + '?format=csv&async=true'
         request = self._build_request(0, -1)
         response = self._cb.post_object(url, request)
-        ref_url = response.get('ref_url', None)
+        ref_url = response.json().get('ref_url', None)
         try:
             job_id = int(ref_url.rsplit('/', 1)[1]) if ref_url else -1
         except ValueError:
--- a/src/cbc_sdk/audit_remediation/base.py
+++ b/src/cbc_sdk/audit_remediation/base.py
@@ -1345,7 +1345,7 @@ class ResultQuery(BaseQuery, QueryBuilderSupportMixin, IterableQueryMixin, Crite
             url += '&download=true'
         request = self._build_request(0, -1)
         self._cb.api_request_stream('POST', url, output, data=request,
-                                     headers={'Accept': 'application/octet-stream' if compressed else 'text/csv'})
+                                    headers={'Accept': 'application/octet-stream' if compressed else 'text/csv'})
 
     def export_csv_as_string(self):
         """
--- a/src/cbc_sdk/audit_remediation/base.py
+++ b/src/cbc_sdk/audit_remediation/base.py
@@ -1344,7 +1344,7 @@ class ResultQuery(BaseQuery, QueryBuilderSupportMixin, IterableQueryMixin, Crite
         if compressed:
             url += '&download=true'
         request = self._build_request(0, -1)
-        self._cb.post_and_get_stream(url, request, output,
+        self._cb.api_request_stream('POST', url, output, data=request,
                                      headers={'Accept': 'application/octet-stream' if compressed else 'text/csv'})
 
     def export_csv_as_string(self):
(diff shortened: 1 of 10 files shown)


emitreva-cb (approved):

src/cbc_sdk/audit_remediation/base.py
```diff
@@ -1408,6 +1409,30 @@ def export_zipped_csv(self, filename):
         with io.open(filename, 'wb') as file:
             self.export_csv_as_stream(file, True)
 
+    def async_export(self):
+        """
+        Create an asynchronous job that exports the results from the run.
```
avanbrunt-cb, line 1414: We should add a note that this is the recommend approach when expecting a large result set

src/cbc_sdk/connection.py
```diff
@@ -488,6 +488,52 @@ def get_raw_data(self, uri, query_parameters=None, default=None, **kwargs):
         else:
             raise ServerError(error_code=result.status_code, message="Unknown error: {0}".format(result.content))
 
+    def get_stream_data(self, uri, stream_output, query_parameters=None, **kwargs):
```
avanbrunt-cb, line 491: This seems like a challenge to maintain in the future as we continue to make more unique functions. Can we determine a better more generic way to combine post_and_get_stream and get_stream_data into a single function? The same goes for iter_lines

src/tests/unit/audit_remediation/test_audit_remediation_base.py
```diff
@@ -309,6 +312,27 @@ def test_result_query_export_file(cbcsdk_mock):
...
+def test_result_query_async_export(cbcsdk_mock, ref_url, func_raises, get_job):
+    """Tests getting a Job from the SDK to asynchronously export a query's results."""
+    cbcsdk_mock.mock_request("POST", "/livequery/v1/orgs/test/runs/run_id/results/_search?format=csv&async=true",
+                             {"ref_url": ref_url})
+    if get_job:
+        cbcsdk_mock.mock_request('GET', '/jobs/v1/orgs/test/jobs/12345', JOB_DETAILS_1)
+    api = cbcsdk_mock.api
+    result_query = api.select(Result).run_id("run_id")
+    with func_raises:
+        job = result_query.async_export()
```
avanbrunt-cb, line 329: What is the intended code that a developer should write for waiting on the job to complete? Should this be managed through an async function where they do results() and it will block or is the expectation that they use a while loop until the status is completed maybe adding a sleep timer to prevent API overload

src/cbc_sdk/platform/jobs.py
```diff
@@ -0,0 +1,223 @@
...
+
+        Args:
+            filename (str): Name of the file to write the results to.
+        """
+        with io.open(filename, 'wb') as file:
+            self.get_output_as_stream(file)
+
+    def get_output_as_lines(self):
+        """
+        Export the results from the job, returning the data as iterated lines of text.
```
avanbrunt-cb, line 129: Can we add a note that this is intended for csv or text files and should not be used for json files. I would expect that we see more and more file types being used for async jobs

src/cbc_sdk/audit_remediation/base.py
abowersox-cb: I'm adding this as a documentation change.

src/cbc_sdk/connection.py
abowersox-cb: I really don't think we're going to have to support any HTTP verbs other than GET and POST for stream output like this.

src/tests/unit/audit_remediation/test_audit_remediation_base.py
abowersox-cb: In my little piece of test code, I did something like this:
```
progress_data = job.get_progress()
while progress_data[1] < progress_data[0]:
    print("Waiting for job to complete...")
    progress_data = job.get_progress()
```
You could throw a sleep() in the middle of that if you wanted.  Actually, I found that I had to put a small sleep() in _before_ doing this, as the progress call would fail if I tried calling it _immediately_ after creating the job.

src/cbc_sdk/platform/jobs.py
abowersox-cb: Adding that as a documentation change.

src/cbc_sdk/connection.py
avanbrunt-cb: This sets a precedence to create a function for each unique verb + mechanism in the future. I'd rather see us follow a model more similar to `api_json_request`. This would help reduce the amount of functions from 4 to 2 request_iter_lines and request_stream. This concern stems from making too many functions which will lead to more confusion over time around what to use. This is always a balance between making a single function do too much and making too many functions however in this case it's the same end result with a slightly different request structure.

src/tests/unit/audit_remediation/test_audit_remediation_base.py
avanbrunt-cb: This is a decent amount of complexity which we could build into a python future option. Can we add that as an option here?

src/cbc_sdk/connection.py
abowersox-cb: Well, I call the functions `api_request_stream` and `api_request_iterate`, but that's what's been done.

The function mocks are done in a clever fashion.  Instead of passing 'GET_STREAM' or 'GET_LINES` to `mock_request`, you pass in `STREAM:GET` or `ITERATE:GET`. Same thing with POST requests.  It required minimal modification to mocking the test code.

(And no, we can't combine those further.  One of them takes a stream parameter to write output, the other is an iterator that has to be called with `yield from`. That makes them about as similar as cheese and chalk.)

avanbrunt-cb: This makes sense and I agree it would not be feasible to combine the functions further

src/tests/unit/audit_remediation/test_audit_remediation_base.py
abowersox-cb: Yes, via the new `Job` method I've just added, `await_completion()`. It returns a `Future` similar to an async query.

src/cbc_sdk/platform/jobs.py
```diff
@@ -84,6 +85,45 @@ def get_progress(self):
...
+            if sleep_time > 0.0:
+                if sleep_time < 1.0:
+                    sleep_time *= 2.0
+            else:
+                sleep_time = 0.1
+        return self
+
+    def await_completion(self):
+        """
+        Waits for this job to complete in an asynchronous fashion.
```
avanbrunt-cb, line 116: I would clarify this description that the function returns a future which provides a non blocking mechanism to fetch the results when they are ready. This function does not block

src/cbc_sdk/platform/jobs.py
```diff
@@ -84,6 +85,45 @@ def get_progress(self):
...
+        progress_data = (1, 0, '')
+        sleep_time = 0.0
+        while progress_data[1] < progress_data[0]:
+            if sleep_time > 0.0:
+                time.sleep(sleep_time)
+            try:
+                progress_data = self.get_progress()
+            except ServerError:
+                progress_data = (1, 0, '')
+            if sleep_time > 0.0:
```
avanbrunt-cb, line 107: Can you add a comment here that the sleep time progressively doubles until it hits 1s where it maintains a 1s sleep time. This will provide clear and immediate understanding of the logic in the future

avanbrunt-cb (approved):
Two minor documentation based comments otherwise everything looks good to me

src/cbc_sdk/platform/jobs.py
abowersox-cb: "The sleep time starts at 0.1 seconds and doubles each time until it hits 1.6 seconds, providing a simple exponential backoff between progress calls."

src/cbc_sdk/platform/jobs.py
abowersox-cb: "Waits for this job to complete in an asynchronous fashion.  Returns a Future object which can be used to await results that are ready to fetch. This function call does not block."

avanbrunt-cb: This phrasing to me is misleading `Waits for this job to complete `. How about this instead? `Create a python future to check for job completion and return results when available`

abowersox-cb: "Python" and "Future" need to be capitalized, but otherwise I'll use that.

src/cbc_sdk/platform/jobs.py
avanbrunt-cb: Hmm. I see your point about how 0.8 will double and be greater than 1.0 though as a user I would ask why the arbitrary 1.6s? Why not keep it simple and use a flat 0.5s or 1s delay? If possible why not make this configurable? My intention for requesting a comment here is to provide context so in the future we understand any reasoning for why we added this mechanism for delay. Should this be a helper function to provide consistent exponential backoff?

abowersox-cb: We're leaving the issue of exponential backoff for another time and changing it to a flat 0.5s delay for now.

PR #269, opened 2022-02-07
[WIP] Fixed formatting of read the docs

kebringer-cb:
Work in progress - some generation errors still to be investigated.

Changes so far:
Added regex to strip ... when copying examples
Updates to format of examples by changing indenting and whitespace
Fixed sphinx formatting warnings
Used this as a reference:
https://www.sphinx-doc.org/en/master/usage/
                     extensions/example_google.html

## Pull request checklist

- [ ] Yes
- [ ] No

### DIFF
--- a/src/cbc_sdk/base.py
+++ b/src/cbc_sdk/base.py
@@ -567,7 +567,9 @@ class NewBaseModel(object, metaclass=CbMetaModel):
 
         Notes:
             Should be overridden for any class wanting to use subobject reporting capabilities.
+
             If the attribute is a list of subobjects, this method should return that list.
+
             If the attribute is a single subobject, this method should return that subobject.
         """
         return None
@@ -1345,11 +1347,11 @@ class QueryBuilder(object):
     through the CBCloudAPI.select API.
 
     Examples:
-    >>> from cbc_sdk.base import QueryBuilder
-    >>> # build a query with chaining
-    >>> query = QueryBuilder().where(process_name="malicious.exe").and_(device_name="suspect")
-    >>> # start with an initial query, and chain another condition to it
-    >>> query = QueryBuilder(device_os="WINDOWS").or_(process_username="root")
+        >>> from cbc_sdk.base import QueryBuilder
+        >>> # build a query with chaining
+        >>> query = QueryBuilder().where(process_name="malicious.exe").and_(device_name="suspect")
+        >>> # start with an initial query, and chain another condition to it
+        >>> query = QueryBuilder(device_os="WINDOWS").or_(process_username="root")
 
     """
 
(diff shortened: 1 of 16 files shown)


src/cbc_sdk/base.py
```diff
@@ -58,11 +58,11 @@ def __new__(mcs, name, bases, clsdict):
             model_data = yaml.safe_load(
                 open(os.path.join(mcs.model_base_directory, swagger_meta_file), 'rb').read())
 
-        clsdict["__doc__"] = "Represents a %s object in the Carbon Black server.\n\n" % (name,)
-        for field_name, field_info in iter(model_data.get("properties", {}).items()):
-            docstring = field_info.get("description", None)
-            if docstring:
-                clsdict["__doc__"] += ":ivar %s: %s\n" % (field_name, docstring)
+        # clsdict["__doc__"] = "Represents a %s object in the Carbon Black server.\n\n" % (name,)
```
kebringer-cb, line 61: Would really like confirmation that removing these lines hasn't broken anything.  It hasn't from what I could see, but could be unexpected side affects.  Commented pending agreement it can really be removed.

emitreva-cb (approved):

src/cbc_sdk/base.py
abowersox-cb: I'm not sure what's going on here.  It looks like the metaclass is trying to b0rk around with the class docstring based on information in the Swagger metafile. `:ivar` is a Sphinx-format documentation directive. If this is to be reused, we should determine how instance-variable documentation works in Google-format doc strings and do that.

abowersox-cb: I've replaced that chunk of code with something that expresses the intent of what that original code was trying to do, but does it following the Google format, and also preserves any existing docstring that was already in the code.  Try generating the docs yourself and let me know what you think.

src/cbc_sdk/base.py
```diff
@@ -2124,28 +2138,36 @@ def _check_range(self, range):
...
 
         Returns:
             Query (AsyncQuery): The Query object that will receive the specified range(s).
 
-        Note: The range parameter must be in this dictionary format:
+        Note:
+            The range parameter must be in this dictionary format:
+
             {
+
```
avanbrunt-cb, line 2154: Why the extra new lines here?

avanbrunt-cb (approved):

src/cbc_sdk/base.py
kebringer-cb: Without the lines it prints as: 
“bucket_size”: “<object>”, “start”: “<object>”, “end”: “<object>”, “field”: “<string>”

instead of each on a new line

PR #270, opened 2022-02-10
CBAPI-3379: Add Device Facet API to CBC SDK

abowersox-cb:
## Pull request checklist

https://jira.carbonblack.local/browse/CBAPI-3379

## Pull Request Description

Added the facets() function to DeviceSearchQuery, which performs query faceting.  The facet values have an extra method of their own, query_devices(), which starts a search query for all devices matching that facet.

- [ ] Yes
- [x] No

## How Has This Been Tested?

Unit tests have been updated and augmented to cover the new functionality. Coverage on that source file is at 94%. A UAT test for facets has also been added and proven to work against PROD02.

### DIFF
--- a/src/cbc_sdk/platform/devices.py
+++ b/src/cbc_sdk/platform/devices.py
@@ -293,8 +293,15 @@ class DeviceFacet(UnrefreshableModel):
             """
             Set up a device query to find all devices that match this facet value.
 
+            Example:
+                >>> facets = api.select(Device).where('').facets(['policy_id'])
+                >>> for value in facets[0].values_:
+                ...     print(f"Policy ID = {value.id}:")
+                ...     for dev in value.query_devices():
+                ...         print(f"    {dev.name} ({dev.last_external_ip_address})")
+
             Returns:
-                DeviceQuery: The new DeviceQuery, which may have additional criteria added to it.
+                DeviceQuery: A new DeviceQuery set with the criteria, which may have additional criteria added to it.
             """
             query = self._cb.select(Device)
             if self._outer.field == 'policy_id':
@@ -755,6 +762,12 @@ class DeviceSearchQuery(BaseQuery, QueryBuilderSupportMixin, CriteriaBuilderSupp
     def facets(self, fieldlist, max_rows=0):
         """
         Return information about the facets for all known evices, using the defined criteria.
+
+        Example:
+            >>> query = api.select(Device).where('')
+            >>> facets = query.facets(['policy_id', 'status', 'os', 'ad_group_id'])
+            >>> for f in facets:
+            ...     print(f"Field {f.field} - {len(f.values_)} distinct values")
 
         Required Permissions:
             device (READ)
--- a/src/cbc_sdk/platform/devices.py
+++ b/src/cbc_sdk/platform/devices.py
@@ -765,7 +765,6 @@ class DeviceSearchQuery(BaseQuery, QueryBuilderSupportMixin, CriteriaBuilderSupp
         result = resp.json()
         return [DeviceFacet(self._cb, None, item) for item in result['results']]
 
-
     def download(self):
         """
         Uses the query parameters that have been set to download all device listings in CSV format.
--- a/src/cbc_sdk/platform/devices.py
+++ b/src/cbc_sdk/platform/devices.py
@@ -13,11 +13,11 @@
 
 """Model and Query Classes for Platform Devices"""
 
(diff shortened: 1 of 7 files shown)


emitreva-cb (approved):

src/tests/uat/platform_devices_uat.py
```diff
@@ -85,6 +99,30 @@ def search_devices(cb):
...
+        f"Test failed: API call returned {len(api_results)}, but SDK call returned {len(list_facets)}"
+
+    for api_facet, list_facet in zip(api_results, list_facets):
+        assert api_facet['field'] == list_facet.field, \
+            f"Test failed: field name from API {api_facet['field']} differs from SDK {list_facet.field}"
+        assert len(api_facet['values']) == len(list_facet.values_), \
+            f"Test failed: length of values for field {list_facet.field} differs - " \
+            f"API says {len(api_facet['values'])} but SDK says {len(list_facet.values_)}"
+        for vdict, val in zip(api_facet['values'], list_facet.values_):
+            assert vdict == val._info, \
```
gyanitsa-cb, line 120: I'm wondering if it is possible for the API to return the values in a different order in the 2 calls. If it is, this will fail due to different order even if as a whole the responses contain the same values.

abowersox-cb: The ordering appears to be consistent between calls. It's primarily "most popular" (highest total value) first.

src/cbc_sdk/platform/devices.py
```diff
@@ -220,6 +255,104 @@ def get_vulnerabilties(self):
...
+            """
+            query = self._cb.select(Device)
+            if self._outer.field == 'policy_id':
+                query.set_policy_ids([int(self.id)])
+            elif self._outer.field == 'status':
+                query.set_status([self.id])
+            elif self._outer.field == 'os':
+                query.set_os([self.id.upper()])
+            elif self._outer.field == 'ad_group_id':
+                query.set_ad_group_ids([int(self.id)])
```
avanbrunt-cb, line 314: Can we add a catch all here if the backend supports additional facet options in the future that way this code doesn't need updated every time a new property is supported? We should be able to check the term name and add it to the criteria object directly using `_update_criteria`

src/cbc_sdk/platform/devices.py
```diff
@@ -616,13 +759,55 @@ def _run_async_query(self, context):
...
+            fieldlist (list[str]): List of facet field names. Valid names are "policy_id", "status", "os",
+                                   and "ad_group_id".
+            max_rows (int): The maximum number of rows to return. 0 means return all rows.
+
+        Returns:
+            list[DeviceFacet]: A list of facet information.
+        """
+        if not fieldlist:
+            raise ApiError("At least one term field must be specified")
+        if not all((field in DeviceSearchQuery.VALID_FACET_FIELDS) for field in fieldlist):
```
avanbrunt-cb, line 785: I understand the need to protect against the potential API error however this would block if the backend API supports additional facet fields. I believe we can rely on the API error to provide this feedback instead of building it in client side

src/cbc_sdk/platform/devices.py
abowersox-cb: This would make sense, but there's not a one-to-one correspondence between the field names as used by faceting and the field names in the criteria.  Some map straight across (`status` => `status`, `os` => `os`), some map to their plurals (`policy_id` => `policy_ids`, `ad_group_id` => `ad_group_ids`).

abowersox-cb: Also, there needs to be conversion done on some fields.  The value as reported in the facet is always a string, but `policy_ids` and `ad_group_ids` require integers.  And `os` is reported by faceting as lower case, but the criteria for it demands upper case.

src/cbc_sdk/platform/devices.py
abowersox-cb: Also, other "facet" APIs like for USB devices use client-side validation of the facet fields in the same fashion.

src/cbc_sdk/platform/devices.py
avanbrunt-cb: Fair point. I did not account for the change in property names

src/cbc_sdk/platform/devices.py
avanbrunt-cb: Given the change in field names this doesn't make sense

PR #281, opened 2022-04-12
CBAPI-3574: Add NSX Remediation to SDK

abowersox-cb:
## Pull request checklist

https://jira.carbonblack.local/browse/CBAPI-3574

## Pull Request Description

Added the NSX Remediation functionality to the SDK.  This includes a new `NSXRemediationJob` object and additions to the `Device` object.

- [ ] Yes
- [x] No

## How Has This Been Tested?

Unit tests have been written with 100% code coverage of all added code.  A UAT script has also been added and has been tested against the dev01 server test organization.

### DIFF
--- a/src/cbc_sdk/platform/devices.py
+++ b/src/cbc_sdk/platform/devices.py
@@ -273,7 +273,7 @@ class Device(PlatformModel):
         Returns:
             set[str]: All current NSX tags that are set for this workload.
         """
-        if self.nsx_available:
+        if self.nsx_available and self._info['nsx_distributed_firewall_policy']:
             return set([tag_name for tag_name in NSXRemediationJob.VALID_TAGS
                         if tag_name in self._info['nsx_distributed_firewall_policy']])
         return set()
--- a/src/cbc_sdk/platform/devices.py
+++ b/src/cbc_sdk/platform/devices.py
@@ -265,22 +265,9 @@ class Device(PlatformModel):
         """
         return self._info['deployment_type'] == 'WORKLOAD' and self._info['nsx_enabled']
 
-    @property
-    def nsx_tags(self):
-        """
-        Returns the current NSX tag(s) set for this workload.
-
-        Returns:
-            set[str]: All current NSX tags that are set for this workload.
-        """
-        if self.nsx_available and self._info['nsx_distributed_firewall_policy']:
-            return set([tag_name for tag_name in NSXRemediationJob.VALID_TAGS
-                        if tag_name in self._info['nsx_distributed_firewall_policy']])
-        return set()
-
     def nsx_remediation(self, tag, set_tag=True):
         """
-        Start an NSX Remediation job on this device.
+        Start an NSX Remediation job on this device to change the tag.
 
         Required Permissions:
             appliances.nsx.remediation(EXECUTE)
@@ -291,11 +278,22 @@ class Device(PlatformModel):
             set_tag (bool): True to toggle the specified tag on, False to toggle it off. Default True.
 
         Returns:
-            NSXRemediationJob: The object representing all running jobs.
+            NSXRemediationJob: The object representing all running jobs.  None if the operation is a no-op.
         """
-        if self.nsx_available:
-            return NSXRemediationJob.start_request(self._cb, self.id, tag, set_tag)
-        raise ApiError("NSX actions are not available on this device")
+        if not self.nsx_available:
+            raise ApiError("NSX actions are not available on this device")
+        current = self._info['nsx_distributed_firewall_policy']
+        if current is None:
+            if not set_tag:
+                return None  # clearing tag is a no-op if no tag is set
+        elif current == tag:
+            if set_tag:
+                return None  # setting tag is a no-op if already set
+        else:
+            if set_tag:
+                raise ApiError(f"NSX tag already set to {current}, cannot set another tag without clearing it")
+            return None  # clearing tag is a no-op in this case
+        return NSXRemediationJob.start_request(self._cb, self.id, tag, set_tag)
 
 
 class DeviceFacet(UnrefreshableModel):
--- a/src/cbc_sdk/platform/devices.py
+++ b/src/cbc_sdk/platform/devices.py
@@ -18,6 +18,7 @@ from cbc_sdk.platform import PlatformModel
 from cbc_sdk.platform.vulnerability_assessment import Vulnerability, VulnerabilityQuery
 from cbc_sdk.base import (UnrefreshableModel, BaseQuery, QueryBuilder, QueryBuilderSupportMixin,
                           CriteriaBuilderSupportMixin, IterableQueryMixin, AsyncQueryMixin)
+from cbc_sdk.workload import NSXRemediationJob
 
 import time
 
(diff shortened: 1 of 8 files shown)


src/cbc_sdk/workload/nsx_remediation.py
```diff
@@ -0,0 +1,139 @@
...
+
+    @property
+    def status(self):
+        """
+        Returns the current status.
+
+        Returns:
+            dict: The current status, mapping individual job IDs to status value dicts.
+        """
+        return self._status
```
avanbrunt-cb, line 139: If the user calls this before using either of the `await_result` functions then the _status will be all None. Should we call `_poll_status` once before returning. We could even check if _running_jobs > 0 such that we don't need to poll if it's already finished

src/cbc_sdk/workload/nsx_remediation.py
```diff
@@ -0,0 +1,139 @@
...
+        Required Permissions:
+            appliances.registration(READ)
+
+        Returns:
+            Future: A future representing the job and its results.
+        """
+        return self._cb._async_submit(lambda arg, kwarg: arg[0].await_result(), self)
+
+    @property
+    def status(self):
```
avanbrunt-cb, line 132: How can I get the list of devices where the NSX remediation failed?

src/tests/uat/nsx_remediation_uat.py
```diff
@@ -0,0 +1,157 @@
...
+
+    Args:
+        device (Device): The device under test.
+
+    Returns:
+        bool: True if the test completed (possibly with errors printed), False if the device was not suitable for test.
+    """
+    print(f"Beginning test on device #{device.id}")
+    before_tag = device.nsx_distributed_firewall_policy
+    if before_tag:
```
avanbrunt-cb, line 110: Can you add some comments indicating what is being performed? I can't easily tell what is being set versus unset?

src/tests/uat/nsx_remediation_uat.py
```diff
@@ -0,0 +1,157 @@
...
+
+from cbc_sdk.helpers import build_cli_parser, get_cb_cloud_object
+from cbc_sdk.platform import Device
+from cbc_sdk.workload import NSXRemediationJob
+from cbc_sdk.errors import ApiError, ServerError, NSXJobError
+
+
+def get_devices(cb, ids):
+    """
+    Gets the devices that can be tested with NSX Remediation.
```
avanbrunt-cb, line 36: Can you specify what indicates that a device can be tested?

src/tests/unit/fixtures/workload/mock_nsx_remediation.py
```diff
@@ -0,0 +1,110 @@
...
+    "action_type": "NSX_REMEDIATION",
+    "options": {
+        "toggle": "ON",
+        "tag": "CB-NSX-Quarantine"
+    }
+}
+
+NSX_RESPONSE_1 = {
+    "job_ids": [
+        "7ff05537-350a-420c-bfa8-3408ac70ce53"
```
avanbrunt-cb, line 14: Use a sample ID fetched from a CBC environment as I believe they don't use a raw guid. Please replace all job_ids below

src/cbc_sdk/workload/nsx_remediation.py
```diff
@@ -0,0 +1,139 @@
...
+import time
+from cbc_sdk.errors import ApiError, ServerError, NSXJobError
+
+
+class NSXRemediationJob:
+    """An object that runs and monitors an NSX Remediation operation."""
+    VALID_TAGS = ['CB-NSX-Quarantine', 'CB-NSX-Isolate', 'CB-NSX-Custom']
+    RUNNING_STATUSES = ['UNASSIGNED', 'SCHEDULED', 'RUNNING', 'RUNNING_UNDELIVERED']
+
+    def __init__(self, cb, running_jobs):
```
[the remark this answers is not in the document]
abowersox-cb: Changing it to `running_job_ids` to be even _more_ specific.

src/cbc_sdk/workload/nsx_remediation.py
abowersox-cb: We call `_poll_status` at the end of the `__init__` method, so by the time the object is constructed, it's already been polled once.

src/cbc_sdk/workload/nsx_remediation.py
abowersox-cb: You can get the list of job IDs that failed by examining `status`, but I know of no way to correlate the job IDs back to devices.  We don't get enough information back when starting a request to establish that.

src/tests/uat/nsx_remediation_uat.py
abowersox-cb: It's basically all devices for which `nsx_available` is `True`. That property being `True` means that the `deployment_type` is `WORKLOAD` and the `nsx_enabled` flag is set.  A comment has been added to that effect.

src/cbc_sdk/platform/devices.py
```diff
@@ -254,6 +255,46 @@ def get_vulnerabilties(self):
...
+        Returns:
+            bool: True if NSX actions are available, False if not.
+        """
+        return self._info['deployment_type'] == 'WORKLOAD' and self._info['nsx_enabled']
+
+    def nsx_remediation(self, tag, set_tag=True):
+        """
+        Start an NSX Remediation job on this device to change the tag.
+
+        Required Permissions:
```
dganev-cb, line 272: I think that should be specified in a different way otherwise it looks like something (python related) is related to use the method. Maybe just a single sentence like `The required permissions for the API are appliances.nsx.remediation(EXECUTE)`.

src/cbc_sdk/platform/devices.py
```diff
@@ -254,6 +255,46 @@ def get_vulnerabilties(self):
...
+        Args:
+            tag (str): The NSX tag to apply to this device. Valid values are "CB-NSX-Quarantine",
+                       "CB-NSX-Isolate", and "CB-NSX-Custom".
+            set_tag (bool): True to toggle the specified tag on, False to toggle it off. Default True.
+
+        Returns:
+            NSXRemediationJob: The object representing all running jobs.  None if the operation is a no-op.
+        """
+        if not self.nsx_available:
+            raise ApiError("NSX actions are not available on this device")
```
dganev-cb, line 284: Quotes are inconsistent during all of the commits. I suspect that pre-commit is skipped. That also happens in `[nsx_remediation.py](https://github.com/carbonblack/carbon-black-cloud-sdk-python/pull/281/files#diff-466b2e4755f09b5752ef75bea324ba1ae06c6d52fc63684d004096bc4d7d42fd)` new additions.

abowersox-cb: What do you mean by this? Also, your pre-commit stuff doesn't work on Windows.

dganev-cb: `'` and `"` are used at random there is no consistency. Also what does not work on pre-commit? I just installed it and used it without any problems on Windows. Eventually you can report an issue https://github.com/pre-commit/pre-commit here.

abowersox-cb: OK, you're using some weird kind of pre-commit thing that I wasn't previously aware of.

abowersox-cb: It says it's installed.

src/tests/unit/fixtures/workload/mock_nsx_remediation.py
abowersox-cb: The job IDs generated are all in the form `NSX_REMEDIATE_{guid}`.  I have changed the job IDs used in the mocks and unit tests to match this pattern.

src/cbc_sdk/workload/nsx_remediation.py
avanbrunt-cb: That makes sense was curious if there was a possibility but we aren't given a mapping if multiple job_ids are returned

avanbrunt-cb (approved):
Looks good.

src/cbc_sdk/platform/devices.py
dganev-cb: I still see the deflake commits it seems to me that it's still not working. I don't understand the comment about the tool being weird but is there some error that you are getting? I would like to know so that I can raise an issue for it, thanks!

abowersox-cb: Is there any way to run the quote-checker independent of the pre-commit mechanism, the way you can run flake8 independent of the pre-commit mechanism?

dganev-cb: There is no quote-checker in the `pre-commit` configuration. I just stated that as inconstancy, I think that we need to agree on a double/single quote strings and be consistent with it. 

If we stick to single quote we can add the `double-quote-string-fixer` step in the pre-commit configuration that would fix the double quoted stuff on each modified file in a commit.

dganev-cb: Resolving this topic since it is not in the scope of the PR.

dganev-cb (approved):

PR #284, opened 2022-04-18
CBAPI-3372 Remove py36 add suport for py39 and py310

dganev-cb:
## Pull request checklist

CBAPI-3372

## Pull Request Description
Remove python3.6 since it's EOL, and add support for 3.9/3.10

- [x] No (All of our clients should be advised to upgrade their python version)

## How Has This Been Tested?
CI


emitreva-cb (approved):

avanbrunt-cb (commented):
What's the reason we should stop supporting python 3.6?

dganev-cb:
> What's the reason we should stop supporting python 3.6?

@avanbrunt-cb Python 3.6 has reached EOL December 2021 and it won't be provided with bug fixes and security patches

dganev-cb:
@avanbrunt-cb @abowersox-cb  the test `test_create_session_with_init_poll_timeout` still connects to the real example.com website despite the mock request. If the tests are run within the folder they are fine but in the `/unit/` scope it connects to the real thing. I suspect that the mock is not being changed or it happens something funky in `pytest` and the fixture has to reset before running that test. Probably has to do something with `scope="function"` but I will keep researching it.

avanbrunt-cb (approved):

abowersox-cb (approved):

PR #290, opened 2022-06-09
CBAPI - 3869 - Enhance the watchlist-feeds-reports guide

dganev-cb:
## Pull request checklist

- [x] Docs have been reviewed and added/updated if needed (for bug fixes/features)

CBAPI-3869

## Pull Request Description
I wanted to create some quick snippets that you can just take and run since I found myself digging and creating those all over again, so I've put them into the guide. I can see that some of the code is being duplicated and I am not sure how to structure it but whenever I look for something IOC-related I just scroll down to the IOC section of that document (I believe the other people are doing the same thing). I am open to suggestions on how to structure the whole thing.

### DIFF
--- a/docs/watchlists-feeds-reports.rst
+++ b/docs/watchlists-feeds-reports.rst
@@ -191,6 +191,94 @@ Both IOCs and reports may include a ``link`` property, which is used by the Carb
 to indicate that this IOC or report is being managed outside of the console.  If this property is not ``None``,
 the console UI will disable the ability to edit the IOC or report, but they can still be edited via the API.
 
+Creating an IOC
++++++++++++++++
+
+You can create an IOC via the `IOC_V2` class, there are 3 avaliable methods that you can use to initiate your IOC:
+`IOC_V2.create_query, IOC_V2.create_equality, IOC_V2.create_regex`.
+
+Creating an equality IOC
+++++++++++++++++++++++++
+
+.. code-block:: python
+
+    >> from cbc_sdk import CBCloudAPI
+    >> from cbc_sdk.enterprise_edr import IOC_V2
+    >> cbcsdk = CBCloudAPI(profile="default")
+    >> IOC_V2.create_equality(cbcsdk, None, "netconn_domain", ["localhost.local"])
+    <cbc_sdk.enterprise_edr.threat_intelligence.IOC_V2: id ad361179-d586-4c99-af3e-821224cc0fd9> @ https://<CBCInstanceURL>
+
+
+Creating a query IOC
+++++++++++++++++++++
+
+.. code-block:: python
+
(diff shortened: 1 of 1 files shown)


docs/watchlists-feeds-reports.rst
```diff
@@ -191,6 +191,94 @@ Both IOCs and reports may include a ``link`` property, which is used by the Carb
...
+.. code-block:: python
+
+    >> from cbc_sdk.enterprise_edr import Watchlist, Report, IOC_V2
+    >> ioc_id = "<ioc_id>"
+    >> report_id = "<report_id>"
+    >> watchlist = cbcsdk.select(Watchlist, "<watchlist_id>")
+    <cbc_sdk.enterprise_edr.threat_intelligence.Watchlist: id <watchlist_id>> @ https://<CBCInstanceURL>
+    >> ioc = IOC_V2.create_regex(cbcsdk, None, "process_name", r"(^/usr/.*$)|(^/bin/.*$)")
+    >> reports = watchlist.reports
+    >> report = [report_ for report_ in reports if report_.id == report_id][0]
```
avanbrunt-cb, line 276: Hmm. This seems odd that we don't have a way to fetch the report directly for a watchlist as the API is available without the watchlist_id. I'd expect something like `cb.select(Report, "id", from_watchlist=True)`

avanbrunt-cb: Any thoughts on creating a story to implement this behavior?

avanbrunt-cb (approved):

docs/watchlists-feeds-reports.rst
dganev-cb: Yes, I also think it's odd that this is not being supported however, you need to provide the Watchlist ID as well as the Report ID to select it so more like the current behavior with the `feed_id`.
```python
cb.select(Report, "id", watchlist_id="...")
```

avanbrunt-cb: Why do you need the Watchlist id? or is that because of the CBC SDK side?

dganev-cb: If we don't have an `watchlist_id` or `feed_id` how do we know where to look for that report?

dganev-cb: Actually just saw the `from_watchlist` which is set to a boolean when the real value should be a string. I will have to check the logic more thoughtfully to create a proper solution for this.

emitreva-cb (approved):

dganev-cb:
Not merging until the release of the new version, also I have to add the new parts about the Report in https://github.com/carbonblack/carbon-black-cloud-sdk-python/pull/292

PR #297, opened 2022-07-18
CBAPI-3881: Policy Simplification

abowersox-cb:
## Pull request checklist

https://jira.carbonblack.local/browse/CBAPI-3641, https://jira.carbonblack.local/browse/CBAPI-3881

## Pull Request Description

New `Policy` object under `cbc_sdk.platform` that implements the new Policy API and yet uses methods and properties that remain backwards-compatible with the old `Policy` object under `cbc_sdk.endpoint_standard` (which has been removed).

- [x] Yes
- [ ] No

The big change is that the new APIs require "custom" API keys with the right permissions, in keeping with the newer API scheme. Permission requirements are documented in the docstrings.  There is a "forwarding import" in `cbc_sdk.endpoint_standard` so basic linkage of the object won't break.

## How Has This Been Tested?

Full unit tests have been written for the new policy code (coverage 98%). In addition, @kebringer-cb has run the new policy objects against the existing policy example, with updated API keys, and all functions work, proving the backwards compatibility.


src/cbc_sdk/platform/policies.py
```diff
@@ -0,0 +1,1222 @@
...
+            model_unique_id (int): ID of the policy.
+            initial_data (dict): Initial data used to populate the policy.
+            force_init (bool): If True, forces the object to be refreshed after constructing.  Default False.
+            full_doc (bool): If True, object is considered "fully" initialized. Default False.
+        """
+        super(Policy, self).__init__(cb, model_unique_id=model_unique_id, initial_data=initial_data,
+                                     force_init=force_init if initial_data else True, full_doc=full_doc)
+        self._object_rules = None
+        self._object_rules_need_load = True
+        self._info["version"] = 2
```
avanbrunt-cb, line 72: We can't override this property as this is a value used in the API otherwise everytime someone uses the SDK it overrides their previous value

src/cbc_sdk/platform/policies.py
```diff
@@ -0,0 +1,1222 @@
...
+        super(PolicyQuery, self).__init__(None)
+        self._doc_class = doc_class
+        self._cb = cb
+        self._count_valid = False
+        self._total_results = 0
+        self._policy_ids = []
+        self._system = False
+        self._system_set = False
+        self._names = []
+        self._descrs = []
```
avanbrunt-cb, line 1008: Can we make this `_descriptions` as `_decsrs` was not obvious as an acryonm

src/tests/unit/platform/test_policies.py
```diff
@@ -0,0 +1,522 @@
...
+    called_full_get = False
+
+    def on_get(uri, query_params, default):
+        nonlocal called_full_get
+        called_full_get = True
+        return FULL_POLICY_1
+
+    cbcsdk_mock.mock_request('GET', '/policyservice/v1/orgs/test/policies/65536', on_get)
+    api = cbcsdk_mock.api
+    policy = Policy(api, 65536, SUMMARY_POLICY_1, False, False)
```
avanbrunt-cb, line 86: Shouldn't we only need to provide the following to get the policy using the class directly?

```
Policy(api, 65536)

or

cb.select(Policy, 65536)
```

src/cbc_sdk/platform/policies.py
```diff
@@ -0,0 +1,1222 @@
+#!/usr/bin/env python3
```
kebringer-cb, line 1: The Policy class doesn't appear in read the docs.
I've configured rtd to build of this branch, so changes can be checked here:
https://carbon-black-cloud-python-sdk.readthedocs.io/en/cbapi-3641_policy_simplification

the file docs/cbc_sdk.platform.rst needs this section added (alphabetically):

cbc\_sdk.platform.policies module
----------------------------------

.. automodule:: cbc_sdk.platform.policies
   :members:
   :undoc-members:
   :show-inheritance:

src/cbc_sdk/platform/policies.py
```diff
@@ -0,0 +1,1222 @@
...
+            self.touch(True)
+
+    class PolicyBuilder:
+        """
+        Builder object to simplify the creation of new Policy objects.
+
+        To use, call Policy.create() to get a PolicyBuilder, use its methods to set the properties of the
+        new policy, call its build() method to build the populated Policy, then call the policy save() method.
+        The org.policy(CREATE) permission is required.
+
```
kebringer-cb, line 83: Somewhere in this section include something like
For information on values for policy settings including enumeration values, see the Policy Service API page:
https://developer.carbonblack.com/reference/carbon-black-cloud/platform/latest/policy-service/#fields

src/cbc_sdk/platform/policies.py
```diff
@@ -0,0 +1,1222 @@
...
+            Returns:
+                PolicyBuilder: This object.
+
+            Raises:
+                ApiError: If the sensor setting name is not a valid one.
+            """
+            if name in Policy.VALID_SENSOR_SETTINGS:
+                self._sensor_settings[name] = value
+            else:
+                raise ApiError(f"invalid sensor setting: {name}")
```
kebringer-cb, line 465: Why does this raise an ApiError, but others return InvalidObjectError?

kebringer-cb (commented):
Example on the policy builder is great.

src/cbc_sdk/platform/policies.py
abowersox-cb: The new version of the policy APIs doesn't actually use the `version` property. It's in there for backwards compatibility reasons.

avanbrunt-cb: Understood however it could be used by a customer and their own tracking purposes

src/cbc_sdk/platform/policies.py
abowersox-cb: Done.

src/tests/unit/platform/test_policies.py
abowersox-cb: This isn't looking up a policy, it's creating one locally for use as part of the test.

avanbrunt-cb: It's creating a policy object with that id in order to fetch the content of that from the Carbon Black Cloud otherwise this would need a POST instead of GET. I believe the above code is equivalent to what you have there currently.

src/cbc_sdk/platform/policies.py
abowersox-cb: Got that, thanks.  I should probably also check for anything that needs removing from `cbc_dsk.endpoint_standard.rst`.

src/cbc_sdk/platform/policies.py
abowersox-cb: Actually, that may be wrong...

src/tests/unit/platform/test_policies.py
abowersox-cb: Line 88 does not "get" a policy. It is _simulating_ a `Policy` object initialized only from summary information, as, for example, would be done by `PolicyQuery`.  The objective here is to show that reading a property that was already initialized from summary information does _not_ trigger a full GET, but reading a property that was _not_ so initialized will trigger a full GET.

src/cbc_sdk/platform/policies.py
abowersox-cb: OK, the code now no longer _overwrites_ `version`, but it adds it if it's not there.

avanbrunt-cb (approved):
The backwards compatibility looks good to me and the break down between Policy and Rules makes sense. I'm a little confused by the function name `object_ruless()` but couldn't think of an alternative give we have `rules()` for backwards compatibility. 

Let's make sure the UAT test covers some of the backwards compability usescases

kebringer-cb (approved):

abowersox-cb:
> Let's make sure the UAT test covers some of the backwards compability usescases

@kebringer-cb conducted a good backwards-compatibility UAT case; she ran the existing policies example script (`examples/endpoint_standard/policy_operations.py`) against the new policy code, using updated API keys.  This was done before the PR was created.

emitreva-cb (approved):

PR #306, opened 2022-08-10
Differential Analysis

hkaragitliev-cb:
## Pull request checklist

## Pull Request Description
Tests are yet to be added in the next sprint

https://jira.carbonblack.local/browse/CBAPI-3183
https://developer.carbonblack.com/reference/carbon-black-cloud/cb-liveops/latest/differential-analysis-api/#rate-limits

- [ ] Yes
- [x] No


emitreva-cb (approved):

docs/differential-analysis.rst
```diff
@@ -0,0 +1,165 @@
...
+**1. Prerequisites**
+
+To perform a Differential Analysis, create the "point-in-time" snapshots of your endpoints with Live Query or use existing ones.
+You can find a step-by-step Live Query API guide `here <https://developer.carbonblack.com/reference/carbon-black-cloud/cb-liveops/latest/livequery-api/#quick-start>`_ and
+a version for the CBC Python SDK `here <https://carbon-black-cloud-python-sdk.readthedocs.io/en/latest/live-query/>`_.
+The example Live Query runs look for added or removed Firefox add-ons.
+
+**2. Query Comparison**
+
+Start a Query Comparison with the id's you received from step 1. In some cases, the ``older_run_id`` is not required.
```
abowersox-cb, line 22: What happens if you don't supply `older_run_id`? Does it just assume "previous run to the run specified by `newer_run_id` "? This may be covered in the Dev Network docs, but it would be nice to see mention of it here.

src/cbc_sdk/audit_remediation/differential.py
```diff
@@ -0,0 +1,293 @@
...
+            raise ApiError("Invalid newer_run_id")
+        self._newer_run_id = newer_run_id
+        return self
+
+    def older_run_id(self, older_run_id):
+        """
+        This can be optional.
+
+        If not specified, the previous run as compared to the primary will be chosen if
+        it is a reccuring one. If comparing two individual runs, this is required.
```
abowersox-cb, line 126: Typo: "reccuring" -> "recurring"

abowersox-cb:
Just a few things to address, from my point of view.  Some I agreed with Kylie on.

docs/differential-analysis.rst
```diff
@@ -0,0 +1,165 @@
+Differential Analysis
+=====================
+
+Differential Analysis provides the ability to compare and understand the changes between two
+`Live Query <https://developer.carbonblack.com/reference/carbon-black-cloud/cb-liveops/latest/livequery-api/>`_ runs.
```
avanbrunt-cb, line 5: Why not link to the Live Query Run class in the CBC SDK?

docs/differential-analysis.rst
```diff
@@ -0,0 +1,165 @@
...
+
+Using the ``.count_only()`` method with a value of ``False`` will allow you to see the actual changes between the two snapshots.
+To use this method, append it to the rest of the Differential object query. The actual changes will be in the ``changes`` property, under ``diff_results``.
+
+    >>> from cbc_sdk import CBCloudAPI
+    >>> from cbc_sdk.audit_remediation import Differential
+    >>>
+    >>> cbc = CBCloudAPI(profile='sample')
+    >>>
+    >>> resp = cbc.select(Differential).newer_run_id('jcdqsju4utpaayj5dh5r2llzffeolg0u').older_run_id('yhbg3wcea9y1l4asiltky5tupkgauzas')
```
avanbrunt-cb, line 85: Does this actually make an API call? or is the `resp` variable still the query object? If the variable is still the query object then we should call it `query`

avanbrunt-cb: Please apply the same logic to all examples

docs/differential-analysis.rst
```diff
@@ -0,0 +1,165 @@
...
+    >>> export = resp.count_only(False).set_device_ids([12345]).async_export()
+    >>> print(export.status)
+    IN_PROGRESS
+    >>> # wait for it to finish and refresh the information in the SDK
+    >>> job_future = export.await_completion()
+    >>> finished_job = job_future.result()
+    >>> finished_job.refresh()
```
avanbrunt-cb, line 160: Why do we need to refresh the finished_job as we did `await_completion()` and `result()`?

src/cbc_sdk/audit_remediation/differential.py
```diff
@@ -0,0 +1,293 @@
...
+            from_row (int): Not used; retained for compatibility.
+            max_rows (int): Not used; retained for compatibility.
+
+        Returns:
+            Iterable: The iterated query.
+        """
+        url = self._build_url()
+        request = self._build_request()
+        results = self._cb.post_object(url, body=request).json()
+        result = results.get("results")[0]
```
avanbrunt-cb, line 246: Will we only have one differential in the results? If so we should add a comment explaining this for now in case we need to modify this behavior in the future

src/cbc_sdk/audit_remediation/models/differential.yaml
```diff
@@ -0,0 +1,23 @@
+type: object
```
avanbrunt-cb, line 1: This doesn't properly reflect the _info object of the Differential class instead this represents the query request json. This needs modified to represent singular result of the response

src/cbc_sdk/audit_remediation/differential.py
```diff
@@ -0,0 +1,293 @@
...
+
+
+class Differential(NewBaseModel):
+    """
+    Represents a Differential Analysis run.
+
+    Example:
+        >>> run = cb.select(Differential).newer_run_id(newer_run_id)
+        >>> print(*run)
+        >>> print(resp[0]._info)
```
avanbrunt-cb, line 36: Our examples should not use _info instead we should show how to get one of the properties like `diff_results`

```
print(resp[0])
print(resp[0].diff_results)
```

src/cbc_sdk/audit_remediation/differential.py
```diff
@@ -0,0 +1,293 @@
...
+    def _count(self):
+        """
+        Returns the number of results from the run of this query.
+
+        Returns:
+            int: The number of results from the run of this query.
+        """
+        return self._total_results
+
+    def _perform_query(self, from_row=None, max_rows=None):
```
avanbrunt-cb, line 232: We should implement a helper class which makes it clearer to get the synchronous results instead of relying on `query[0]` given at the moment we only get one result back. We could use `def submit()` as we used a similar experience for Live Query Runs https://carbon-black-cloud-python-sdk.readthedocs.io/en/latest/live-query/#start-a-query-run

docs/guides-and-resources.rst
```diff
@@ -29,6 +29,7 @@ Guides
 * :doc:`alerts` - Work and manage different types of alerts such as CB Analytics Alert, Watchlist Alerts and Device Control Alerts.
 * :doc:`device-control` - Control the blocking of USB devices on endpoints.
 * :doc:`live-query` - Live Query allows operators to ask questions of endpoints
+* :doc:`differential-analysis` - Provides the ability to compare and understand the changes between two Live Query runs
```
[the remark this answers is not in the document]
hkaragitliev-cb: Everywhere else, we advertise it as a separate product (that's how the PMs wanted it), so I think just move it up so we get the proper alphabetical order.

docs/differential-analysis.rst
hkaragitliev-cb: It is covered in the dev net, but nice catch - adding it.
How it works - If it is a recurring Live Query run, it will automatically get previous to the supplied one. If you supply just a single run query id, the API will throw a specific error.

docs/differential-analysis.rst
hkaragitliev-cb: Something went very wrong here, most of the stuff is not needed, thanks!

src/cbc_sdk/audit_remediation/differential.py
hkaragitliev-cb: The result is just one always, but there can be multiple objects in the diff_results. All the differentials for multiple devices are there:
`
{
  "org_key": "4JDT3MX9Q",
  "results": [
    {
      "newer_run_id": "w6zhfxwgr9vujflop0rdqj7krtl09vlr",
      "newer_run_create_time": "2022-09-27T19:13:46.731Z",
      "older_run_id": "7obm6dhnjxuo35ahcmglcpnjpfuoucsb",
      "older_run_create_time": "2022-09-27T18:56:38.133Z",
      "diff_processed_time": 0.049,
      "newer_run_not_responded_devices": [
        17331059
      ],
      "older_run_not_responded_devices": [
        17331059
      ],
      "diff_results": [
        {
          "device_id": 16801738,
          "change_count": 1,
          "added_count": 1,
          "removed_count": 0,
          "changes": [
            {
              "action": "ADDED",
              "fields": [
                {
                  "key": "name",
                  "value": "Facebook Container"
                }
              ]
            }
          ],
          "older_run_row_count": 17,
          "newer_run_row_count": 18
        },
        {
          "device_id": 11412673,
          "change_count": 2,
          "added_count": 2,
          "removed_count": 0,
          "changes": [
            {
              "action": "ADDED",
              "fields": [
                {
                  "key": "name",
                  "value": "Disconnect"
                }
              ]
            },
            {
              "action": "ADDED",
              "fields": [
                {
                  "key": "name",
                  "value": "HTTPS Everywhere"
                }
              ]
            }
          ],
          "older_run_row_count": 22,
          "newer_run_row_count": 24
        }
      ]
    }
  ]
}
`

src/cbc_sdk/audit_remediation/differential.py
hkaragitliev-cb: I used the _perform_query to maintain a consistent experience with the SDK. I think most of the queries return iterators?
I think the submit() will add more complexity for the user - they will have to read the docs to perform the query, where if they receive an iterator, it should be straight-forward for a python user what they need to do with it.
If this is a strong preference, I'll change it do use the `def submit()`

avanbrunt-cb: I'm stuck on the developer experience given the iterator of the query will only contain one result which is the differential object that contains these properties
```
"newer_run_id": "qpopjb82whlmthlo0x1wrwcnoyxmrueu",
"newer_run_create_time": "2022-05-11T10:49:36.190Z",
"older_run_id": "kibloccplynombvigcgtu2et2zayhzal",
"older_run_create_time": "2022-05-11T10:47:48.952Z",
"diff_processed_time": 0.072,
"newer_run_not_responded_devices": [],
"older_run_not_responded_devices": [],
```

What the developer will likely focus on is the `diff_results` which we can't directly return because the above properties may still be useful.

We already have the following pattern for LiveQuery which uses `.submit()` instead of the query iterator mechanism
```
>>> query_string = "SELECT filename, path FROM file WHERE path = 'C:\\Windows\\Temp\\dbutil_2_3.sys\\' OR path LIKE 'C:\\Users\\%\\AppData\\Local\\Temp\\dbutil_2_3.sys';"
>>> query_object = api.select(Run).where(sql=query_string)
>>> run = query_object.submit()
```

With submit we can return the object directly and avoid needing to return the results list plus if we wanted we could add an async optional parameter to avoid needing another function for `async_export`

The only drawback I see is if in the future they support a multiple run differential but that seems unlikely at the moment given the performance issues but this should be verified before making the final decision

hkaragitliev-cb: Thanks for the feedback; I added the submit() functionality.

avanbrunt-cb (approved):

src/cbc_sdk/audit_remediation/differential.py
avanbrunt-cb: Thanks for the changes this looks much cleaner

hkaragitliev-cb:
Coverage fails because of a broken license. Overriding it.

PR #322, opened 2023-01-09
Add verification for User.create()

avanbrunt-cb:
## Pull request checklist

/CBAPI-4325

## Pull Request Description

Add catch no `User.create` for valid CBCloudAPI object

- [ ] Yes
- [x] No

## How Has This Been Tested?

Yes existing and new unit tests pass

### DIFF
--- a/src/cbc_sdk/platform/users.py
+++ b/src/cbc_sdk/platform/users.py
@@ -17,6 +17,9 @@ from cbc_sdk.errors import ApiError, ServerError, ObjectNotFoundError
 from cbc_sdk.platform.grants import Grant, normalize_org
 import time
 import copy
+import logging
+
+log = logging.getLogger(__name__)
 
 """User Models"""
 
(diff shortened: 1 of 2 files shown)


src/cbc_sdk/platform/users.py
```diff
@@ -290,6 +293,12 @@ def create(cls, cb, template=None):
             UserBuilder: If template is None, returns an instance of this object. Call methods on the object to set
                          the values associated with the new user, and then call build() to create it.
         """
+        try:
+            if cb.__class__.__name__ != 'CBCloudAPI':
+                raise Exception
+        except:
+            raise ApiError("Unable to create User without CBCloudAPI")
```
abowersox-cb, line 300: I was about to flag the use of this as an unnecessary raise-and-catch, but I forgot that `cb` could be `None` or an object type that doesn't have a `__class__` attribute.  Instead, I wonder if this should be made into a static method somewhere (perhaps a `@classmethod` on `CBCloudAPI` itself?) so this check can be easily added to other similar functions.

avanbrunt-cb: I wasn't really sure what to return in a static method beyond the name but that is already feasible with this and we still run into the same problem of an exception being raised so I decided that didn't bring any additional value given we don't have other API objects which would benefit from a class identifying function

emitreva-cb (approved):

abowersox-cb (approved):

emitreva-cb (approved):

PR #333, opened 2023-03-06
Cbapi 4452 network threat metadata

emitreva-cb:
## Pull request checklist

## Pull Request Description
Adding network threat meta data lookup (tms_rule_id is not currently returned from Observations, but will be)

- [ ] Yes
- [ ] No

### DIFF
--- a/src/cbc_sdk/platform/observations.py
+++ b/src/cbc_sdk/platform/observations.py
@@ -14,6 +14,7 @@
 from cbc_sdk.base import UnrefreshableModel, NewBaseModel, FacetQuery
 from cbc_sdk.base import Query
 from cbc_sdk.errors import ApiError, TimeoutError, InvalidObjectError
+from cbc_sdk.platform.network_threat_metadata import NetworkThreatMetadata
 
 import logging
 import time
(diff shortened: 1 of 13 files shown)


src/cbc_sdk/platform/observations.py
```diff
@@ -173,6 +173,14 @@ def _get_detailed_results(self):
                 message="user-specified timeout exceeded while waiting for results"
             )
 
+    def get_network_threat_metadata(self):
+        """Requests Network Threat Metadata.
+
+        Returns:
+            NetworkThreatMetadata: Get the metadata for a given detector (rule).
+        """
+        return NetworkThreatMetadata(self._cb, self.rule_id)
```
avanbrunt-cb, line 182: I believe this may be `tms_rule_id` instead of `rule_id`. Though need to confirm with real data

avanbrunt-cb: Note not every observation type will support this look up only observations related to NTA and IDS. We may actually only want to do this on alerts instead of observations where `tms_rule_id` is definitely available

src/cbc_sdk/platform/observations.py
```diff
@@ -601,3 +610,52 @@ def __getitem__(self, item):
         else:
             raise AttributeError("'{0}' object has no attribute '{1}'".format(self.__class__.__name__,
                                                                               item))
+
+
+class NetworkThreatMetadata(NewBaseModel):
+    """Represents an NetworkThreatMetadata"""
```
kebringer-cb, line 616: should be "a Network..." instead of "an ..."

emitreva-cb: Fixed

src/cbc_sdk/platform/observations.py
emitreva-cb: The name is based on real data. Also, does this mean we should move it out of the observations? Should we move it to a separate python module?

avanbrunt-cb: This will be useful for determining if the type of observation or alert will have threat metadata as rule_id will be populated for all types in the future which won't work for threat metadata.

This also needs to be exposed in the Alert class too so it will probably need a separate file to prevent an infinite import loop

avanbrunt-cb (approved):
We will need to make note to add the helper function to the alert v7 changes in the future

hkaragitliev-cb (approved):
LGTM

PR #338, opened 2023-03-09
CBAPI-4376: Add Core Prevention Endpoints to Policy

abowersox-cb:
## Pull request checklist

https://jira.carbonblack.local/browse/CBAPI-4376

## Pull Request Description

Add Core Prevention rule configuration objects to policies. This also involved a substantial rewrite of the previous "basic" rule configuration code to reflect how rule configurations are _actually_ dealt with within policies.

- [ ] Yes
- [x] No

## How Has This Been Tested?

Test coverage is 97% for policies, 95% for policy_ruleconfigs.


src/tests/unit/platform/test_policy_ruleconfigs.py
```diff
@@ -0,0 +1,296 @@
...
+    cbcsdk_mock.mock_request('GET', '/policyservice/v1/orgs/test/policies/65536', FULL_POLICY_1)
+    api = cbcsdk_mock.api
+    policy = Policy(api, 65536, FULL_POLICY_1, False, True)
+    # Replace all rule configs with the base class for purposes of this test
+    cfgs = policy._info.get("rule_configs", [])
+    ruleconfigobjects = [PolicyRuleConfig(api, policy, cfg['id'], cfg, force_init=False, full_doc=True) for cfg in cfgs]
+    policy._object_rule_configs = dict([(rconf.id, rconf) for rconf in ruleconfigobjects])
+    policy._object_rule_configs_need_load = False
+    # proceed with test
+    rule_config = random.choice(list(policy.object_rule_configs.values()))
```
avanbrunt-cb, line 135: I'm not a fan of the use of `random.choice` for testing as this produces potentially inconsistent testing results which may be hard to discover in the future

src/tests/unit/platform/test_policy_ruleconfigs.py
```diff
@@ -0,0 +1,296 @@
...
+    policy._object_rule_configs = dict([(rconf.id, rconf) for rconf in ruleconfigobjects])
+    policy._object_rule_configs_need_load = False
+    # proceed with test
+    with pytest.raises(NotImplementedError):
+        policy.delete_rule_config('88b19232-7ebb-48ef-a198-2a75a282de5d')
+
+
+def test_rule_config_modify_by_base_method_invalid_id(cb):
+    """Tests modifying a PolicyRuleConfig object using replace_rule_config, but with an invalid ID."""
+    policy = Policy(cb, 65536, copy.deepcopy(FULL_POLICY_1), False, True)
```
avanbrunt-cb, line 181: Instead of creating the policy in most of these tests why not make it fixture that resets with each execution?

src/tests/unit/platform/test_policy_ruleconfigs.py
```diff
@@ -0,0 +1,296 @@
...
+        assert body == CORE_PREVENTION_UPDATE_1
+        put_called = True
+        return copy.deepcopy(body)
+
+    cbcsdk_mock.mock_request('GET', '/policyservice/v1/orgs/test/policies/65536/configs/presentation',
+                             POLICY_CONFIG_PRESENTATION)
+    cbcsdk_mock.mock_request('PUT', '/policyservice/v1/orgs/test/policies/65536/rule_configs/core_prevention', on_put)
+    api = cbcsdk_mock.api
+    policy = Policy(api, 65536, copy.deepcopy(FULL_POLICY_1), False, True)
+    rule_config = policy.object_rule_configs['c4ed61b3-d5aa-41a9-814f-0f277451532b']
```
avanbrunt-cb, line 268: This pattern for fetching a specific rule config makes a challenging experience if the user doesn't have the ids of the specific rule configs for that category. Could we make a core_prevention helper function which fetches all the core_prevention configs at the policy level?

avanbrunt-cb: I think it would be worth considering what the normal usage is going to be for the CBC SDK and the rule configs. How are users going to lookup rule configs? What are they going to do with it once they have the rule config? Let's discuss this in a meeting next week to identify some usage patterns

src/tests/unit/platform/test_policy_ruleconfigs.py
abowersox-cb: This code should work on _any_ arbitrary rule config. I could set it up to always use the "first" one or whatever, but that runs the risk of special characteristics of that _particular_ rule config to cause the test to be passed while masking problems.

src/tests/unit/platform/test_policy_ruleconfigs.py
abowersox-cb: It was just as easy to copy the line of code that creates the policy.  Besides, in some of the tests, we use the `cb` fixture (because no calls are expected to actually be made to the server) and in some, we use `cbcsdk_mock` (because we're making a call to the server that needs a mock response). To do as you say, I'd need two separate fixtures.

src/tests/unit/platform/test_policy_ruleconfigs.py
abowersox-cb: I'm certainly willing to add the helper function at the policy level; it's pretty much a one-liner, as far as I can tell.

src/tests/unit/platform/test_policy_ruleconfigs.py
avanbrunt-cb: However this could also produce random failures which may be hard to diagnose. Please find a more consistent and repeatable testing pattern

src/tests/unit/platform/test_policy_ruleconfigs.py
avanbrunt-cb: Please put this in a fixture to focus the test functions on the desired component that is being tested. The Policy object being created that way is a non standard practice we don't want to draw more attention to either

src/tests/unit/platform/test_policy_ruleconfigs.py
avanbrunt-cb: Please consider the usability of the added core prevention functionaity and how it will be consumed by a developer

src/tests/unit/platform/test_policy_ruleconfigs.py
abowersox-cb: How 'bout I just test _every_ rule configuration in those lists, just to make sure? It's actually almost as easy to write as just picking _one._

src/tests/unit/platform/test_policy_ruleconfigs.py
abowersox-cb: You're probably right that we need to just sit down and talk about this part, because I can't read your mind and see what _you_ mean by "usability...and how it will be consumed by a developer."

src/tests/unit/platform/test_policy_ruleconfigs.py
avanbrunt-cb: That's better as the test would be consistent and repeatable

src/tests/unit/platform/test_policy_ruleconfigs.py
avanbrunt-cb: Here is the desired fixture to be added so we can remove the Policy creation from the test functions
```
@pytest.fixture(scope="function")
def policy(monkeypatch, cb):
    """Mocks CBC SDK for unit tests"""
    return Policy(cb, 65535, copy.deepcopy(FULL_POLICY_1), False, True)
```

abowersox-cb: This change has been pushed.

src/tests/unit/platform/test_policy_ruleconfigs.py
```diff
@@ -0,0 +1,293 @@
...
+    cfgs = policy._info.get("rule_configs", [])
+    ruleconfigobjects = [PolicyRuleConfig(api, policy, cfg['id'], cfg, force_init=False, full_doc=True) for cfg in cfgs]
+    policy._object_rule_configs = dict([(rconf.id, rconf) for rconf in ruleconfigobjects])
+    policy._object_rule_configs_need_load = False
+    # proceed with test
+    for rule_config in policy.object_rule_configs.values():
+        old_name = rule_config.name
+        old_category = rule_config.category
+        old_parameters = rule_config.parameters
+        rule_config.refresh()
```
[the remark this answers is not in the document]
abowersox-cb: Actually, this is the test where we replace all the rule config objects in the policy with base PolicyRuleConfig objects, to test the refresh on the base class. So no, this particular instance is correct.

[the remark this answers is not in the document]
abowersox-cb: We test the core prevention refresh in `test_core_prevention_refresh`, which is on or about line 201.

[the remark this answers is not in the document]
avanbrunt-cb: ah yeah I didn't notice that we manualy created it

avanbrunt-cb (approved):

src/cbc_sdk/platform/policy_ruleconfigs.py
```diff
@@ -0,0 +1,274 @@
+#!/usr/bin/env python3
+
+# *******************************************************
+# Copyright (c) VMware, Inc. 2020-2022. All Rights Reserved.
```
hkaragitliev-cb, line 4: We need to bump the dates in all headers

PR #339, opened 2023-03-14
UAT Observations + improvements based on KK discussion

emitreva-cb:
## Pull request checklist

## Pull Request Description
* UAT for Observations
* static methods for suggestions for both Observation / Alert
* static method for bulk_get_details
* remove validation & suggestions from rest_api.

- [ ] Yes
- [ ] No

### DIFF
--- a/src/cbc_sdk/platform/alerts.py
+++ b/src/cbc_sdk/platform/alerts.py
@@ -252,6 +252,29 @@ class BaseAlert(PlatformModel):
         """
         return self._update_threat_workflow_status("OPEN", remediation, comment)
 
+    @staticmethod
+    def search_suggestions(cb, query):
+        """
+        Returns suggestions for keys and field values that can be used in a search.
+
+        Args:
+            cb (CBCloudAPI): A reference to the CBCloudAPI object.
+            query (str): A search query to use.
+
+        Returns:
+            list: A list of search suggestions expressed as dict objects.
+
+        Raises:
+            ApiError: if cb is not instance of CBCloudAPI
+        """
+        from cbc_sdk.rest_api import CBCloudAPI
+        if not isinstance(cb, CBCloudAPI):
+            raise ApiError("cb argument should be instance of CBCloudAPI.")
+        query_params = {"suggest.q": query}
+        url = "/appservices/v6/orgs/{0}/alerts/search_suggestions".format(cb.credentials.org_key)
+        output = cb.get_object(url, query_params)
+        return output["suggestions"]
+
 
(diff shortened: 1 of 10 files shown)


src/cbc_sdk/platform/alerts.py
```diff
@@ -252,6 +252,29 @@ def update_threat(self, remediation=None, comment=None):
...
+            cb (CBCloudAPI): A reference to the CBCloudAPI object.
+            query (str): A search query to use.
+
+        Returns:
+            list: A list of search suggestions expressed as dict objects.
+
+        Raises:
+            ApiError: if cb is not instance of CBCloudAPI
+        """
+        from cbc_sdk.rest_api import CBCloudAPI
```
emitreva-cb, line 270: If the import is at the beginning it raises an error for circular import.

src/cbc_sdk/platform/alerts.py
```diff
@@ -252,6 +252,29 @@ def update_threat(self, remediation=None, comment=None):
         """
         return self._update_threat_workflow_status("OPEN", remediation, comment)
 
+    @staticmethod
+    def search_suggestions(cb, query):
```
emitreva-cb, line 256: Thought it is a tautology for the name to be `alerts_search_suggestions`, but if you think it is better to use `alerts_search_suggestions` I will change it.

src/cbc_sdk/platform/alerts.py
avanbrunt-cb: The other alternative would be using `obj.__class__.__name__`

src/cbc_sdk/platform/observations.py
```diff
@@ -192,6 +227,59 @@ def get_network_threat_metadata(self):
         except AttributeError:
             raise ApiError("No available network threat metadata.")
 
+    @staticmethod
+    def search_suggestions(cb, query, count=None):
```
emitreva-cb, line 231: Thought it is tautology for the name to be `observations_search_suggestions`, but if you think it is better to use `observations_search_suggestions`

src/cbc_sdk/platform/alerts.py
avanbrunt-cb: `search_suggestions` makes sense as it will be `Alert.search_suggestions`

src/cbc_sdk/platform/alerts.py
emitreva-cb: Thanks, I will change that, but probably tomorrow.

hkaragitliev-cb (approved):

PR #341, opened 2023-03-17
CBAPI-4547 add static concept

emitreva-cb:
## Pull request checklist

- [ ] Yes
- [ ] No

### DIFF
--- a/docs/concepts.rst
+++ b/docs/concepts.rst
@@ -518,7 +518,7 @@ Get details for all events per alert
 Static Methods
 --------------
 
-As of version 1.4.2 we are introducing static methods on some classes. They handle API requests that are not tied with a specific resouce id, thus they cannot be instance methods, but static helper methods. And because those methods are static, you need to pass CBCloudAPI object as a first argument.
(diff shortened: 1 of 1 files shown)


avanbrunt-cb (approved):

docs/concepts.rst
```diff
@@ -513,3 +513,84 @@ Get details for all events per alert
   Category: ['OBSERVED']
   Type: NETWORK
   Alert Id: ['BE084638']
+
+
+Static Methods
+--------------
+
+As of version 1.4.2 we are introducing static methods on some classes. They handle API requests that are not tied with a specific resource id, thus they cannot be instance methods, but static helper methods. And because those methods are static, you need to pass CBCloudAPI object as a first argument.
```
kebringer-cb, line 521: slight rewording:
In version 1.4.2 we introduced static methods on some classes. They handle API requests that are not tied to a specific resource id, thus they cannot be instance methods, instead static helper methods. Because those methods are static, they need a CBCloudAPI object to be passed as the first argument.

emitreva-cb: Thanks, fixed.

PR #342, opened 2023-03-17
Multi org alerts fixed

jclark-vmware:
## Pull request checklist

https://jira.carbonblack.local/browse/CBAPI-4221

## Pull Request Description

Adding sample script for retrieving alerts from multiple orgs

- [ ] Yes
- [x] No

### DIFF
--- a/docs/alerts.rst
+++ b/docs/alerts.rst
@@ -82,11 +82,25 @@ Retrieving Alerts for Multiple Organizations
 
 With the example below, you can retrieve alerts for multiple organizations.
 
-Create a csv file with values that match the profile names in your credentials.cbc file. You can modify the code inside the for loop to fit your needs.
-
 .. code-block:: python
 
- >>> from cbc_sdk import CBCloudAPI
+    >>> from cbc_sdk import CBCloudAPI
+    >>> from cbc_sdk.platform import BaseAlert
+    >>> org_list = ["org1", "org2"]
+    >>> for org in org_list:
+    ...     org = ''.join(org)
+    ...     api = CBCloudAPI(profile=org)
+    ...     alerts = api.select(BaseAlert).set_minimum_severity(7)[:5]
+    ...     print('Results for Org {}'.format(org))
+    >>> for alert in alerts:
+    ...     print(alert.id, alert.device_os, alert.device_name, alert.category)
+    ...
+    ...
+
+
+You can also read from a csv file with values that match the profile names in your credentials.cbc file.
+
+    >>> from cbc_sdk import CBCloudAPI
     >>> from cbc_sdk.platform import BaseAlert
     >>> import csv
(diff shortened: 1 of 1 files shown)


kebringer-cb (approved):

docs/alerts.rst
```diff
@@ -68,6 +68,32 @@ You can also filter on different kind of **TTPs** (*Tools Techniques Procedures*
...
+    >>> file = open ("data.csv", "r", encoding='utf-8-sig')
+    >>> org_list = list(csv.reader(file, delimiter=","))
+    >>> file.close()
+    >>> for org in org_list:
+    ...     org = ''.join(org)
+    ...     api = CBCloudAPI(profile=org)
+    ...     alerts = api.select(BaseAlert).set_minimum_severity(7)[:5]
+    ...     print('Results for Org {}'.format(org))
+    >>> for alert in alerts:
+    ...     print(alerts[0].id, alerts[0].device_os, alerts[0].device_name, alerts[0].category)
```
kebringer-cb, line 92: I think this line can be alert.id etc instead of alert[0] because it's in the for loop.

jclark-vmware: good catch, thank you!

docs/alerts.rst
```diff
@@ -68,6 +68,32 @@ You can also filter on different kind of **TTPs** (*Tools Techniques Procedures*
...
+
+Create a csv file with values that match the profile names in your credentials.cbc file. You can modify the code inside the for loop to fit your needs.
+
+.. code-block:: python
+
+ >>> from cbc_sdk import CBCloudAPI
+    >>> from cbc_sdk.platform import BaseAlert
+    >>> import csv
+    >>> file = open ("data.csv", "r", encoding='utf-8-sig')
+    >>> org_list = list(csv.reader(file, delimiter=","))
```
avanbrunt-cb, line 84: Why use a csv file instead of creating a list of profile names in the code itself?

abowersox-cb: This is just an example, maybe this example actually _looks_ more like a use case one of our customers would have?

avanbrunt-cb: If we are going to use the csv file method then would recommending opening the file using `with:` as this guarantees close if something goes wrong

jclark-vmware: @abowersox-cb you'e correct. I was going for simplicity, because the audience for this script is not very familiar with python. @avanbrunt-cb I'll look into using with:

abowersox-cb:
I'm going to let @jclark-vmware respond to the latest comments by @avanbrunt-cb before I merge this.

docs/alerts.rst
jclark-vmware: Done

avanbrunt-cb (approved):

abowersox-cb (approved):

PR #343, opened 2023-03-20
Updates to setup.py and change one Dockerfile to use setup.py-based install exclusively

abowersox-cb:
## Pull request checklist

N/A

## Pull Request Description

Fix an issue with docs generation by adding a package to `setup.py` in addition to `requirements.txt`. Also, change the RHEL Dockerfile so it does not use `pip3 install -r requirements.txt`, just `pip3 install .` This will test for imbalances between the two setup files.

- [ ] Yes
- [x] No

## How Has This Been Tested?

This PR is being used to test the changes.

### DIFF
--- a/docker/rhel/Dockerfile
+++ b/docker/rhel/Dockerfile
@@ -7,4 +7,4 @@ WORKDIR /app
 RUN dnf install -y redhat-rpm-config gcc libffi-devel python38-devel openssl-devel
 RUN pip3 install --upgrade pip
 #RUN pip3 install -r requirements.txt
-RUN python3 setup.py test
+RUN pip3 install .[test]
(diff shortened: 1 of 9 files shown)


avanbrunt-cb (approved):

docker/rhel/Dockerfile
```diff
@@ -6,5 +6,5 @@ WORKDIR /app
 
 RUN dnf install -y redhat-rpm-config gcc libffi-devel python38-devel openssl-devel
 RUN pip3 install --upgrade pip
-RUN pip3 install -r requirements.txt
+#RUN pip3 install -r requirements.txt
```
avanbrunt-cb, line 9: Looks like the build failed because the dev dependencies need to be specified separately given `pip install .` only installs required ones

abowersox-cb: I put all the dev dependencies in the `tests_requires` list in `setup.py`, we'll see if that works.

abowersox-cb: I'm trying the former. If that doesn't work, we'll try the latter.

avanbrunt-cb (approved):

PR #353, opened 2023-04-27
CBAPI-4378: Add Host Based Firewall Basic Endpoints to Policy

abowersox-cb:
## Pull request checklist

https://jira.carbonblack.local/browse/CBAPI-4378

## Pull Request Description

Added the host-based firewall rule configurations to Policy, with the basic CRUD operations defined, including add/remove operations for firewall rule groups and rules.

- [ ] Yes
- [x] No

## How Has This Been Tested?

Unit tests have been written to cover added code. Coverage on rule configs source file now 96%.

### DIFF
--- a/src/cbc_sdk/platform/policy_ruleconfigs.py
+++ b/src/cbc_sdk/platform/policy_ruleconfigs.py
@@ -302,9 +302,7 @@ class CorePreventionRuleConfig(PolicyRuleConfig):
 
 
 class HostBasedFirewallRuleConfig(PolicyRuleConfig):
-    """
-    Represents a host-based firewall rule configuration in the policy.
-    """
+    """Represents a host-based firewall rule configuration in the policy."""
     swagger_meta_file = "platform/models/policy_ruleconfig.yaml"
 
     def __init__(self, cb, parent, model_unique_id=None, initial_data=None, force_init=False, full_doc=False):
--- a/src/cbc_sdk/platform/policy_ruleconfigs.py
+++ b/src/cbc_sdk/platform/policy_ruleconfigs.py
@@ -272,3 +272,7 @@ class CorePreventionRuleConfig(PolicyRuleConfig):
         if mode not in ("REPORT", "BLOCK"):
             raise ApiError(f"invalid assignment mode: {mode}")
         self.set_parameter("WindowsAssignmentMode", mode)
+
+
+class HostBasedFirewallRuleConfig(PolicyRuleConfig):
+    pass
--- a/src/cbc_sdk/platform/policy_ruleconfigs.py
+++ b/src/cbc_sdk/platform/policy_ruleconfigs.py
@@ -275,4 +275,189 @@ class CorePreventionRuleConfig(PolicyRuleConfig):
 
 
 class HostBasedFirewallRuleConfig(PolicyRuleConfig):
-    pass
+    """
+    Represents a host-based firewall rule configuration in the policy.
+    """
+    swagger_meta_file = "platform/models/policy_ruleconfig.yaml"
+
+    def __init__(self, cb, parent, model_unique_id=None, initial_data=None, force_init=False, full_doc=False):
+        """
+        Initialize the CorePreventionRuleConfig object.
+
+        Args:
+            cb (BaseAPI): Reference to API object used to communicate with the server.
+            parent (Policy): The "parent" policy of this rule configuration.
+            model_unique_id (str): ID of the rule configuration.
+            initial_data (dict): Initial data used to populate the rule configuration.
+            force_init (bool): If True, forces the object to be refreshed after constructing.  Default False.
+            full_doc (bool): If True, object is considered "fully" initialized. Default False.
+        """
+        super(HostBasedFirewallRuleConfig, self).__init__(cb, parent, model_unique_id, initial_data, force_init,
+                                                          full_doc)
+        self._rule_groups = []
+        self._rule_groups_valid = False
+
+    class FirewallRuleGroup(MutableBaseModel):
+        """Represents a group of related firewall rules."""
+        swagger_meta_file = "platform/models/firewall_rule_group.yaml"
+
+        def __init__(self, cb, initial_data):
+            """
+            Initialize the FirewallRuleGroup object.
+
+            Args:
+                cb (BaseAPI): Reference to API object used to communicate with the server.
+                initial_data (dict): Initial data used to populate the firewall rule group.
+            """
+            super(HostBasedFirewallRuleConfig.FirewallRuleGroup, self).__init__(cb, None, initial_data, False, True)
+            self._rules = [HostBasedFirewallRuleConfig.FirewallRule(cb, d) for d in initial_data.get("rules", [])]
+
+        def _flatten(self):
+            """
+            Turns this rule group into a dict for transferral to the server.
+
+            Returns:
+                dict: The information defining the rule group and its constituent rules.
+            """
+            rc = copy.deepcopy(self._info)
+            rc['rules'] = [rule._flatten() for rule in self._rules]
+            return rc
+
+        @property
+        def rules(self):
+            """
+            Returns a list of the firewall rules within this rule group.
+
+            Returns:
+                list(HostBasedFirewallRuleConfig.FirewallRule): List of contained firewall rules.
+            """
+            return self._rules
+
+        def append_rule(self, rule):
+            self._rules.append(rule)
+
+        def remove(self):
+            if self in self.rule_groups:
+                self.rule_groups.remove(self)
+
+    class FirewallRule(MutableBaseModel):
+        """Represents a single firewall rule."""
+        swagger_meta_file = "platform/models/firewall_rule.yaml"
+
+        def __init__(self, cb, initial_data):
+            """
+            Initialize the FirewallRule object.
+
+            Args:
+                cb (BaseAPI): Reference to API object used to communicate with the server.
+                initial_data (dict): Initial data used to populate the firewall rule.
+            """
+            super(HostBasedFirewallRuleConfig.FirewallRule, self).__init__(cb, None, initial_data, False, True)
+
+        def _flatten(self):
+            """
+            Turns this rule into a dict for transferral to the server.
+
+            Returns:
+                dict: The information defining the rule.
+            """
+            return copy.deepcopy(self._info)
+
+        def remove(self):
+            group_list = [group for group in self.rule_groups if self in group._rules]
+            if group_list:
+                group_list[0]._rules.remove(self)
+
+    def _base_url(self):
+        """
+        Calculates the base URL for these particular rule configs, including the org key and the parent policy ID.
+
+        Returns:
+            str: The base URL for these particular rule configs.
+
+        Raises:
+            InvalidObjectError: If the rule config object is unparented.
+        """
+        return super(HostBasedFirewallRuleConfig, self)._base_url() + "/host_based_firewall"
+
+    def _refresh(self):
+        """
+        Refreshes the rule configuration object from the server.
+
+        Required Permissions:
+            org.policies (READ)
+
+        Returns:
+            bool: True if the refresh was successful.
+
+        Raises:
+            InvalidObjectError: If the object is unparented or its ID is invalid.
+        """
+        return_data = self._cb.get_object(self._base_url())
+        ruleconfig_data = [d for d in return_data.get("results", []) if d.get("id", "") == self._model_unique_id]
+        if ruleconfig_data:
+            self._info = ruleconfig_data[0]
+            self._rule_groups = []
+            self._rule_groups_valid = False
+        else:
+            raise InvalidObjectError(f"invalid host-based firewall ID: {self._model_unique_id}")
+        return True
+
+    def _update_ruleconfig(self):
+        """Perform the internal update of the rule configuration object."""
+        ...
+
+    def _delete_ruleconfig(self):
+        """Perform the internal delete of the rule configuration object."""
+        ...
+
+    @property
+    def enabled(self):
+        return self.get_parameter('enable_host_based_firewall')
+
+    def set_enabled(self, flag):
+        self.set_parameter('enable_host_based_firewall', flag)
+
+    @property
+    def default_action(self):
+        default_rule = self.get_parameter('default_rule')
+        return default_rule.get("action", "ALLOW")
+
+    def set_default_action(self, action):
+        if action not in ("ALLOW", "BLOCK"):
+            raise ApiError(f"invalid default action: {action}")
+        default_rule = self.get_parameter('default_rule')
+        default_rule['action'] = action
+        self.set_parameter("default_rule", default_rule)
+
+    @property
+    def rule_groups(self):
+        if not self._rule_groups_valid:
+            rg_param = self.get_parameter("rule_groups")
+            if rg_param is not None:
+                self._rule_groups = [HostBasedFirewallRuleConfig.FirewallRuleGroup(self._cb, d) for d in rg_param]
+            else:
+                self._rule_groups = []
+            self._rule_groups_valid = True
+        return self._rule_groups
+
+    def new_rule_group(self, name, description):
+        return HostBasedFirewallRuleConfig.FirewallRuleGroup(self._cb, {"name": name, "description": description,
+                                                                        "rules": []})
+
+    def append_rule_group(self, rule_group):
+        self.rule_groups.append(rule_group)
+
+    def new_rule(self, action, direction, protocol, remote_ip):
+        if action not in ("ALLOW", "BLOCK", "BLOCK_ALERT"):
+            raise ApiError(f"invalid rule action: {action}")
+        if direction not in ("IN", "OUT", "BOTH"):
+            raise ApiError(f"invalid rule direction: {direction}")
+        if protocol not in ("TCP", "UDP"):
+            raise ApiError(f"invalid rule protocol: {protocol}")
+        return HostBasedFirewallRuleConfig.FirewallRule(self._cb, {"action": action, "application_path": "*",
+                                                                   "direction": direction, "enabled": True,
+                                                                   "protocol": protocol, "local_ip_address": "*",
+                                                                   "local_port_ranges": "*",
+                                                                   "remote_ip_address": remote_ip,
+                                                                   "remote_port_ranges": "*", "test_mode": False})
--- a/src/cbc_sdk/platform/policy_ruleconfigs.py
+++ b/src/cbc_sdk/platform/policy_ruleconfigs.py
@@ -328,7 +328,7 @@ class HostBasedFirewallRuleConfig(PolicyRuleConfig):
         """Represents a group of related firewall rules."""
         swagger_meta_file = "platform/models/firewall_rule_group.yaml"
 
-        def __init__(self, cb, parent, initial_data):
+        def __init__(self, cb, initial_data):
             """
             Initialize the FirewallRuleGroup object.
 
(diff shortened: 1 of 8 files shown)


src/cbc_sdk/platform/policy_ruleconfigs.py
```diff
@@ -272,3 +299,316 @@ def set_assignment_mode(self, mode):
...
+            self._parent = parent
+            for rule in self._rules:
+                rule._parent = parent
+
+        def _field_updated(self, attrname):
+            """Method called whenever a field is updated."""
+            if self._parent:
+                self._parent._mark_changed()
+
+        def _flatten(self):
```
avanbrunt-cb, line 358: Why do we need `_flatten` when the `_info` object contains the updated full content as python uses pass by reference? From what I can tell the location where `_flatten` is used does not modify the content only sends it to the Carbon Black Cloud backend.

src/cbc_sdk/platform/policy_ruleconfigs.py
```diff
@@ -272,3 +299,316 @@ def set_assignment_mode(self, mode):
...
+            Initialize the parent of this group and its rules.
+
+            Args:
+                parent (HostBasedFirewallRuleConfig): The parent to be set.
+            """
+            self._parent = parent
+            for rule in self._rules:
+                rule._parent = parent
+
+        def _field_updated(self, attrname):
```
avanbrunt-cb, line 353: Instead of piping a new function throughout the entire SDK why not override the `__set__` function and embedd the code here then call the super class to maintain the same functionality?

src/cbc_sdk/platform/policy_ruleconfigs.py
```diff
@@ -272,3 +299,316 @@ def set_assignment_mode(self, mode):
...
+            Args:
+                cb (BaseAPI): Reference to API object used to communicate with the server.
+                initial_data (dict): Initial data used to populate the firewall rule group.
+            """
+            super(HostBasedFirewallRuleConfig.FirewallRuleGroup, self).__init__(cb, None, initial_data, False, True)
+            self._parent = None
+            self._rules = [HostBasedFirewallRuleConfig.FirewallRule(cb, d)
+                           for d in initial_data.get("rules", [])]
+
+        def _init_parent(self, parent):
```
avanbrunt-cb, line 342: This should be an optional parameter in the constuctor and not implemented as a separate protected function

src/cbc_sdk/platform/policy_ruleconfigs.py
```diff
@@ -272,3 +299,316 @@ def set_assignment_mode(self, mode):
...
+        def remove(self):
+            """Removes this rule from the rule group that contains it."""
+            if self._parent:
+                group_list = [group for group in self._parent.rule_groups if self in group._rules]
+                if group_list:
+                    group_list[0]._rules.remove(self)
+                    self._parent._mark_changed()
+                    self._parent = None
+
+    def _base_url(self):
```
avanbrunt-cb, line 436: I missed this in the previous rule config PRs. Why are we not implmenenting this as a `urlobject` property on the class? Why does this need to be a function?

src/cbc_sdk/platform/policy_ruleconfigs.py
```diff
@@ -272,3 +299,316 @@ def set_assignment_mode(self, mode):
...
+        if not self._rule_groups_valid:
+            rg_param = self.get_parameter("rule_groups")
+            if rg_param is not None:
+                self._rule_groups = [HostBasedFirewallRuleConfig.FirewallRuleGroup(self._cb, d) for d in rg_param]
+                for group in self._rule_groups:
+                    group._init_parent(self)
+            else:
+                self._rule_groups = []
+            self._rule_groups_valid = True
+        return self._rule_groups
```
avanbrunt-cb, line 560:
```suggestion
        if self._info.get("rule_groups"):
             # includes parent initialization inside rule group constructor
             self._rule_groups = [HostBasedFirewallRuleConfig.FirewallRuleGroup(self._cb, d, self) for d in rg_param]
        return self._rule_groups
```

The additional `_rule_groups_valid` bool seems unnecesssary when we can check with `self._info.get("rule_groups")` as this returns false if the property is not present or if the list is empty

src/tests/unit/platform/test_policy_ruleconfigs.py
```diff
@@ -280,3 +289,265 @@ def on_delete(url, body):
...
+    rule_config = policy.host_based_firewall_rule_config
+    assert not rule_config.enabled
+    assert rule_config.default_action == "ALLOW"
+    groups = rule_config.rule_groups
+    assert len(groups) == 1
+    assert groups[0].name == "Argon_firewall"
+    assert groups[0].description == "Whatever"
+    rules = groups[0].rules_
+    assert len(rules) == 1
+    assert rules[0].action == "ALLOW"
```
avanbrunt-cb, line 309: Instead of verifying each property individually why not use `rule_config._info == HBFW_GET_RESULT["results"][0]` then step into the sub objects and do further step in comparisons like `groups._info == HBFW_GET_RESULT["results"][0]["parameters]["rulesets"]`

This also seems to be doing two tests at once with initial parsing but also a refresh which should use alternative values which would change before and after the refresh. I'd recommend these being different tests

src/cbc_sdk/platform/policy_ruleconfigs.py
abowersox-cb: Believe it or not, I originally _had_ this as a constructor parameter, but split this off as the only instance in which the `_parent` references need setting is when we actually load the contents of `_rule_groups`. I guess I should put that back then.

src/cbc_sdk/platform/policy_ruleconfigs.py
abowersox-cb: I actually _considered_ overriding `_set`, the function in `MutableBaseModel` that does the heavy lifting for `__set__`, but there were too many paths through that function that resulted in the field _not_ being actually updated, and I wanted a clear signal as to when it _did_ actually get updated, to minimize unnecessary round-trips to the server when `save` is called on something that hasn't actually _changed._  The best way to do that, in my opinion, was to add the `_field_updated` "hook" to `_set`. If you've got a better way in which this could be implemented by _just_ overriding `_set` without having the hook, I'm all ears.

src/cbc_sdk/platform/policy_ruleconfigs.py
abowersox-cb: I implemented `_flatten` in both `FirewallRuleGroup` and `FirewallRule` for reasons of symmetry. It can certainly be omitted from `FirewallRule` by just tapping `_info` directly, but it's still needed in `FirewallRuleGroup`.

src/cbc_sdk/platform/policy_ruleconfigs.py
abowersox-cb: Because that base URL always contains the policy ID of the "parent" policy for these rule configs, and the base class implementation (in `PolicyRuleConfig`) reflects that. It's set up so that _only_ the base class implementation of `_base_url` needs to get the parent policy ID value; all the subclass base URLs are just that URL with something else tacked onto the end, as you see here.  It _could_ be written to use `urlobject`, but that wouldn't be very good encapsulation because I'd have to "leak" the parent policy ID into the subclasses.

src/cbc_sdk/platform/policy_ruleconfigs.py
abowersox-cb: The purpose of `_rule_groups_valid` is to signal that the value of `_rule_groups` is, in fact, valid, because the "lazy load" process has been done.  This is a pattern that I've used elsewhere in the SDK. Perhaps it should be named something like `_rule_groups_loaded`?

Aside from that, the change to use a constructor parameter instead of `_init_parent` has already been noted and will be done.

src/cbc_sdk/platform/policy_ruleconfigs.py
```diff
@@ -272,3 +299,316 @@ def set_assignment_mode(self, mode):
...
+        Returns:
+            FirewallRule: The new firewall rule. Append it to a rule group using the group's append_rule method.
+        """
+        if action not in ("ALLOW", "BLOCK", "BLOCK_ALERT"):
+            raise ApiError(f"invalid rule action: {action}")
+        if direction not in ("IN", "OUT", "BOTH"):
+            raise ApiError(f"invalid rule direction: {direction}")
+        if protocol not in ("TCP", "UDP"):
+            raise ApiError(f"invalid rule protocol: {protocol}")
+        return HostBasedFirewallRuleConfig.FirewallRule(self._cb, {"action": action, "application_path": "*",
```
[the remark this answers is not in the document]
abowersox-cb: While I still think we'll need default values in case _nothing_ is specified for any given value, the use of `kwargs` to specify optional fields is a good idea and I'll want to implement that.

src/cbc_sdk/platform/policy_ruleconfigs.py
avanbrunt-cb: `_flatten` is not needed on `FirewallRuleGroup` as it uses pass by reference and already contains all the rules same with `HostBasedFirewallRuleConfig` as this contains all the `FirewallRuleGroup` and `FirewallRule`

src/cbc_sdk/platform/policy_ruleconfigs.py
avanbrunt-cb: That's why I recommended overriding `__set__` instead of `_set` then calling the super function to maintain existing functionality you may need to override `__setattr__` as well

src/cbc_sdk/platform/policy_ruleconfigs.py
avanbrunt-cb: `_parent` already exists in every RuleConfig class as it inherits the `PolicuRuleConfig` class which is why they should all have a unique `urlobject` property `/policyservice/v1/orgs/{0}/policies/{1}/rule_configs/host_based_firewall` where it can be set by using the following code
```
self.urlobject.format(self._cb.credentials.org_key, self._parent._model_unique_id)
```

src/tests/unit/platform/test_policy_ruleconfigs.py
avanbrunt-cb: That's fair though in this case let's not check every property as it leads to bloated tests and is not increasing code coverage

abowersox-cb: Agreed. I can certainly omit a few of those without sacrificing generality.  Same would apply to the post-refresh test, when it gets split out.

src/cbc_sdk/platform/policy_ruleconfigs.py
abowersox-cb: The reason why I don't like that style is because it forces us to reference `self._parent._model_unique_id` at all levels of the class hierarchy, not _just_ in the base class. However, I'm not gonna get all religious about it, and I'll do it your way.

src/cbc_sdk/platform/policy_ruleconfigs.py
abowersox-cb: What does overriding `__set__` (and maybe `__setattr__` as well) get me that just overriding `_set` couldn't, in this case?

Looking at the implementation in `MutableBaseModel`, `__setattr__` has several different code paths it can go through. It can go through the property setter objects (which ultimately hit `_set` themselves), it can go through `_set` directly, or it can go through `object.__setattr__`. We don't worry about that last case because it won't be changing anything that is going to get pushed to the server.  Neither it nor `NewBaseModel` actually override `__set__` themselves, so maybe that can be discounted.

src/cbc_sdk/platform/policy_ruleconfigs.py
abowersox-cb: The purpose of `_flatten` in `FirewallRuleGroup` is to update the _actual_ `rules` property in the new copy of `_info` with the contents of the `FirewallRule` objects. When those objects are created, they are "divorced" from the original data fed in by the server. When we want to push it _back_ to the server, we need to fill that in, or else the changes to the rules aren't going to be reflected properly. This is _especially_ the case if we add or remove a rule on the rule group.

src/cbc_sdk/platform/policy_ruleconfigs.py
avanbrunt-cb: I guess I am confused why your initial override attempt for `_set` didn't work

src/cbc_sdk/platform/policy_ruleconfigs.py
avanbrunt-cb: Maybe I missed it by I don't see use making a copy anywhere so I assumed the reference was still intact even though we are only referencing a sub component of the original json/_info

abowersox-cb: In any event, adds and removes are _definitely_ going to require `_flatten` or something very much like it.  In neither of those cases is `_info["rules"]` actually touched, we just manipulate the `_rules` list. Since that's the case, it's easier to just do a `_flatten` whenever we need to get the changed information.  And, as part of the `save`, the new information is returned, with which we update `_info` anyway and cause the `_rules` to be regenerated whenever the rule groups list is.

avanbrunt-cb: That's not necessary if we implement the adds/removes correctly as it will modify the pass by reference objects/lists

src/cbc_sdk/platform/policy_ruleconfigs.py
abowersox-cb: Hmm...I wonder if an override for `_set` would look like this...
```
def _set(self, attrname, new_value):
    pre_change_value = self._info.get(attrname)
    super(self, classname)._set(attrname, new_value)
    if self._info.get(attrname) != pre_change_value:
        self._mark_changed()
```

Maybe I should try that? It seems a little clunky though.

avanbrunt-cb: Why do you need check if the change actually took place is there an error condition which would prevent the change?

I would expect the following
```
def _set(self, attrname, new_value):
    super(self, classname)._set(attrname, new_value)
    self._mark_changed()
```

abowersox-cb: But then `_mark_changed` would get called even if the value of that field _didn't actually change._ This defeats the whole purpose of doing this, as we would then be making unnecessary round-trips to the server for fields that didn't actually change.

avanbrunt-cb: Why is `_set` getting called on properties which are not being changed?

src/cbc_sdk/platform/policy_ruleconfigs.py
abowersox-cb: This one will be fixed at the next check-in.

src/cbc_sdk/platform/policy_ruleconfigs.py
abowersox-cb: I'm changing that variable to `_rule_groups_loaded` and also simplifying this function quite a bit by making the `get_parameter` method take a default value parameter. This will be in place at the next check-in.

src/cbc_sdk/platform/policy_ruleconfigs.py
abowersox-cb: Because `_set` gets called for _every_ attempt to set a property value. Inside _that_ method (on or about line 840 in `base.py`) is where it checks for an attempt to set the field to its current value, and it bails out if that's the case.  Therefore, we don't know, _a priori,_ whether the value of the property is going to change inside `_set`, unless we check for it ourselves.

src/cbc_sdk/platform/policy_ruleconfigs.py
abowersox-cb: I felt `urlobject_single` was more appropriate to use here, and the code has been rewritten to use it and eliminate `_base_url()` entirely.  This will be in the next commit.

src/tests/unit/platform/test_policy_ruleconfigs.py
abowersox-cb: I split the test and pared down the number of properties that were checked (by about half, I think, in the case of the "contents" test). I was even able to simplify the "contents" test a bit as a result. This will be in the next commit.

src/cbc_sdk/platform/policy_ruleconfigs.py
abowersox-cb: I've modified those objects to override `_set` instead and removed the "hook" function altogether.  The test I use actually makes use of the `_dirty_attributes` dict, which the base class `_set` manages. This reduces the required operations to simple `in`, `not in`, and Boolean tests, and reduces the intermediate storage required to a single `bool` value. I think now I'm at the "as simple as possible, but no simpler" stage. This will be committed today.

src/cbc_sdk/platform/policy_ruleconfigs.py
abowersox-cb: I made this better in the code that was just committed.

src/cbc_sdk/platform/policy_ruleconfigs.py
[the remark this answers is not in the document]
abowersox-cb: This has actually been implemented.

src/tests/unit/platform/test_policy_ruleconfigs.py
```diff
@@ -280,3 +289,276 @@ def on_delete(url, body):
...
+    assert not policy.is_dirty()
+    rule_config = policy.host_based_firewall_rule_config
+    result_groups = [group for group in rule_config.rule_groups if group.name == "Isolate"]
+    assert len(result_groups) == 1
+    result_groups[0].remove()
+    assert policy.is_dirty()
+    rule_config.save()
+    assert put_called
+    result_groups = rule_config.rule_groups
+    assert len(result_groups) == 1
```
avanbrunt-cb, line 563: Why does the length of the group remain 1 if it was removed above? 

Also can we add a test which has multiple rules in a rule group to ensure the correct rule is removed as only 1 rulegroup or rule doesn't test the new add/remove logic

src/cbc_sdk/platform/policy_ruleconfigs.py
avanbrunt-cb: Thanks the new code looks great!

src/cbc_sdk/platform/policy_ruleconfigs.py
```diff
@@ -402,21 +395,14 @@ def _set(self, attrname, new_value):
...
-                dict: The information defining the rule.
-            """
-            return copy.deepcopy(self._info)
-
         def remove(self):
             """Removes this rule from the rule group that contains it."""
             if self._parent:
                 group_list = [group for group in self._parent.rule_groups if self in group._rules]
                 if group_list:
+                    location = [ndx for ndx, item in enumerate(group_list[0]._rules) if item == self]
```
avanbrunt-cb, line 403: Have you considered using `list.index()` instead of the custom one liner loop into a list. The index function does suppport object values as the input. Python reference: https://docs.python.org/3/tutorial/datastructures.html

avanbrunt-cb: This will also be safer as you may get an index out of range error if the object is not found

src/tests/unit/platform/test_policy_ruleconfigs.py
abowersox-cb: > Why does the length of the group remain 1 if it was removed above?

We're measuring two different things. Lines 556-557 are measuring the number of groups which were found matching the name "Isolate." That should always be 1.  Lines 562-563 are measuring the number of groups left _after_ the "Isolate" group is removed. That should also be 1.

> Also can we add a test which has multiple rules in a rule group to ensure the correct rule is removed as only 1 rulegroup or rule doesn't test the new add/remove logic

I thought of this. The definition of the policy in the mock `FULL_POLICY_5` has two rule groups, one of which has two rules, the other of which has only one.  It was added for the specific purpose of testing rule addition and removal.  The structure of the rule groups in `FULL_POLICY_5` is as follows:
* Group `Crapco_firewall`
  * Rule `my_first_rule`
  * Rule `DoomyDoomsOfDoom`
* Group `Isolate`
  * Rule `BlockIRC`

src/cbc_sdk/platform/policy_ruleconfigs.py
abowersox-cb: OK, I'm rewriting this to use `index()` instead. In this context, it will _not_ fail, though, since we already know `self` exists in `group_list[0].rules`, because that was why the group got selected in the search expression that created `group_list`.

.readthedocs.yaml
```diff
@@ -0,0 +1,34 @@
...
+    python: "3.8"
+    # You can also specify other tool versions:
+    # nodejs: "19"
+    # rust: "1.64"
+    # golang: "1.19"
+
+# Build documentation in the docs/ directory with Sphinx
+sphinx:
+  configuration: docs/conf.py
+# fail_on_warning: true
```
avanbrunt-cb, line 21: Can we make a new story to re-enable fail on warnings or associate with the ticket if we have one to fix the warningss already

avanbrunt-cb (approved):

.readthedocs.yaml
abowersox-cb: @kebringer-cb does a ticket already exist for fixing all the warnings in that build?
(Kylie told me that some of those doc fixes are not easy, she's looked at them.)

kebringer-cb: There is now: CBAPI-4705

emitreva-cb (approved):

PR #359, opened 2023-05-23
CBAPI-4379: Add Advanced Host-Based Firewall Endpoints to Policy

abowersox-cb:
## Pull request checklist

https://jira.carbonblack.local/browse/CBAPI-4379

## Pull Request Description

Add support for the "copy HBFW rules to policy" and "export HBFW rules" endpoints to the `HostBasedFirewallRuleConfig`.

- [ ] Yes
- [x] No

## How Has This Been Tested?

Unit tests written for all new functions. Test coverage at 97%.

### DIFF
--- a/src/cbc_sdk/platform/policy_ruleconfigs.py
+++ b/src/cbc_sdk/platform/policy_ruleconfigs.py
@@ -560,7 +560,7 @@ class HostBasedFirewallRuleConfig(PolicyRuleConfig):
         self._mark_changed()
         return rule_group
 
-    def copy_rules_to(self, *args):
+    def copy_rules(self, *args):
         """
         Copies the parameters for host-based firewall rule configurations to another policy or policies.
 
--- a/src/cbc_sdk/platform/policy_ruleconfigs.py
+++ b/src/cbc_sdk/platform/policy_ruleconfigs.py
@@ -612,4 +612,7 @@ class HostBasedFirewallRuleConfig(PolicyRuleConfig):
             raise ApiError(f"Invalid format: {format}")
         url = self.urlobject_single.format(self._cb.credentials.org_key, self._parent._model_unique_id)\
             + "/rules/_export"
-        return self._cb.get_raw_data(url, {"format": format})
+        if format == "json":
+            return self._cb.get_object(url, {"format": format})
+        else:
+            return self._cb.get_raw_data(url, {"format": format})
--- a/src/cbc_sdk/platform/policy_ruleconfigs.py
+++ b/src/cbc_sdk/platform/policy_ruleconfigs.py
@@ -559,3 +559,57 @@ class HostBasedFirewallRuleConfig(PolicyRuleConfig):
             self._info['parameters']['rule_groups'] = [rule_group._info]
         self._mark_changed()
         return rule_group
+
+    def copy_rules_to(self, *args):
+        """
+        Copies the parameters for host-based firewall rule configurations to another policy or policies.
+
+        Required Permissions:
+            org.firewall.rules(UPDATE)
+
+        Args:
+            args (list[Any]): References to policies to copy to. May be Policy objects, integers, or other.
+
+        Returns:
+            dict: Result structure from copy operation.
+
+        Raises:
+            ApiError: If the parameters could not be converted to policy IDs.
+        """
+        from cbc_sdk.platform.policies import Policy
+        target_ids = []
+        for arg in args:
+            if isinstance(arg, Policy):
+                target_ids.append(arg.id)
+            elif isinstance(arg, int):
+                target_ids.append(arg)
+            else:
+                try:
+                    target_ids.append(int(str(arg)))
+                except ValueError:
+                    raise ApiError(f"invalid policy ID or reference: {arg}")
+        if not target_ids:
+            raise ApiError("at least one policy ID or reference must be specified")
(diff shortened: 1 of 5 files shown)


src/cbc_sdk/platform/policy_ruleconfigs.py
```diff
@@ -559,3 +559,60 @@ def append_rule_group(self, name, description):
...
+
+    def copy_rules_to(self, *args):
+        """
+        Copies the parameters for host-based firewall rule configurations to another policy or policies.
+
+        Required Permissions:
+            org.firewall.rules(UPDATE)
+
+        Args:
+            args (list[Any]): References to policies to copy to. May be Policy objects, integers, or other.
```
avanbrunt-cb, line 571: I'm a little confused by other and how this makes the function a little confusing. I would also rename the function to `copy_rules` the suffix `_to` seems a little confusing

src/cbc_sdk/platform/policy_ruleconfigs.py
```diff
@@ -559,3 +559,60 @@ def append_rule_group(self, name, description):
...
+        Raises:
+            ApiError: If the parameters could not be converted to policy IDs.
+        """
+        from cbc_sdk.platform.policies import Policy
+        target_ids = []
+        for arg in args:
+            if isinstance(arg, Policy):
+                target_ids.append(arg.id)
+            elif isinstance(arg, int):
+                target_ids.append(arg)
```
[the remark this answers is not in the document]
abowersox-cb: I think we can leave the validation to the server side in this case.

src/cbc_sdk/platform/policy_ruleconfigs.py
abowersox-cb: OK, here's how it's supposed to work. If you pass in a Policy, it grabs the ID. If you pass in an integer, it uses that as a policy ID. If you pass anything else, it tries to convert it to a string, then to an integer. This may throw `ValueError`, which becomes an `ApiError`. I will rename the function though.

src/cbc_sdk/platform/policy_ruleconfigs.py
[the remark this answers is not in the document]
avanbrunt-cb: Can you confirm the behavior for me as to what happens?

src/cbc_sdk/platform/policy_ruleconfigs.py
avanbrunt-cb: I understand what it's trying to do but I don't understand why we implemented `other`. Do we have any legitmate usecases to expect something that needs converted to a string then to an integer as any of our class models would not work

abowersox-cb: Well, maybe we pass in a policy ID that's actually a string equivalent of a number, like "12345678". The test case for this function demonstrates this.

src/tests/unit/platform/test_policy_ruleconfigs.py
```diff
@@ -601,3 +605,68 @@ def on_put(url, body, **kwargs):
...
+        assert body == HBFW_COPY_RULES_PUT_REQUEST
+        put_called = True
+        return copy.deepcopy(HBFW_COPY_RULES_PUT_RESPONSE)
+
+    cbcsdk_mock.mock_request('PUT', '/policyservice/v1/orgs/test/policies/1492/rule_configs/host_based_firewall/_copy',
+                             on_put)
+    api = cbcsdk_mock.api
+    policy = Policy(api, 1492, copy.deepcopy(FULL_POLICY_5), False, True)
+    target_policy = Policy(api, 65536, copy.deepcopy(FULL_POLICY_1), False, True)
+    result = policy.host_based_firewall_rule_config.copy_rules_to(601, target_policy, "344")
```
abowersox-cb, line 625: @avanbrunt-cb This is the test case that shows the various kinds of input to `copy_rules_to` (soon to be `copy_rules`).

src/cbc_sdk/platform/policy_ruleconfigs.py
avanbrunt-cb: The point I'm trying to make is we should be more specific and not say other. Policy object or policy_id in string or int form. The concept of other leads to confusion on what we support

src/tests/unit/platform/test_policy_ruleconfigs.py
avanbrunt-cb: Understood. I'm more referring to our doc string and mentioning other.

src/cbc_sdk/platform/policy_ruleconfigs.py
avanbrunt-cb: I think that's for you to verify and determine. What happens if a boolean is provided or an Alert? What happens if an invalid policy id is provided? Does the error or response provide enough context to fix or code against? I'm asking for the doc string to be updated to remove other as for the code as long as it matches the doc string and returns a consumable response then I am happy

abowersox-cb: Well, the doc string has been updated. The ApiError thrown as a result of ValueError includes the actual argument that failed conversion to an integer, so if it's an object that caused the problem, you'll see what object it is. As for invalid policy IDs, I still need to try the API in Postman, as you asked in the comments below.

avanbrunt-cb (approved):
Approved assuming there isn't a weird documentation issue on the export too

emitreva-cb (approved):

PR #381, opened 2023-07-28
Adjust read the docs

avanbrunt-cb:
## Pull Request Description

I modified the maxdepth of the SDK Documentation toctree to 4 as this enables the left nav to have the abiliity to see the file contents which makes identifiying and switching to the desired function easier

Screenshot 2023-07-28 at 11 51 23 AM

Additionally I think we want to swap  `:inherited-members:` for `:undoc-members:` as the undoc members seem to mostly be the properties which get pulled in from the swagger file and documented twice. The swagger properties are documented under the models Parameters section which should be enough. Plus any of the properties we do want users to see should have doc strings so there should be no undoc-members which we actually want documented

Screenshot 2023-07-28 at 11 54 41 AM

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -9,7 +9,7 @@ version: 2
 build:
   os: ubuntu-22.04
   tools:
-    python: "3.8"
+    python: "3.11"
     # You can also specify other tool versions:
     # nodejs: "19"
     # rust: "1.64"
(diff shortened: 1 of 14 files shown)


kebringer-cb (approved):

jclark-vmware (approved):
Love it!

PR #394, opened 2023-09-01
CBAPI-3765: Refresh RTD User Guide Sections

abowersox-cb:
## Pull request checklist

https://jira.carbonblack.local/browse/CBAPI-3765

## Pull Request Description

Certain pages in the RTD Guide for CBC SDK have been updated: Installation, Authentication, Getting Started, Guides, Concepts _(removed entirely),_ new Guide for Searching, left-hand sidebar _(collapse the Guides sub-listing)._

- [ ] Yes
- [x] No

## How Has This Been Tested?

Changes have been inspected locally.


docs/porting-guide.rst
```diff
@@ -52,7 +59,6 @@ Import statements will need to change:
     # note that the original "Event" has been decommissioned
     from cbc_sdk.endpoint_standard import Device, EnrichedEvent, Policy
```
kebringer-cb, line 60: While we're updating, could we change this to use the platform Device and Policy classes, and also a note to consider using Observations instead of Enriched Events.   Blog about the benefits of Observations is here to add a bit of context or reference something: https://developer.carbonblack.com/2023/07/how-to-take-advantage-of-the-new-observations-api/.  One of the other benefits is that they apply to all products, not only endpoint standard.

from cbc_sdk.endpoint_standard import EnrichedEvent
from cbc_sdk.platform import Device, Policy

docs/getting-started.rst
```diff
@@ -19,10 +22,12 @@ See also the :doc:`installation` section of this documentation for more informat
...
 --------------
+
 To make use of APIs, you will need an *API token,* in case you are using Carbon Black Cloud to manage your
-identity and authentication, or if you are using VMware Cloud Services Platform, an *OAuth App with Bearer* or a *Personal API Token*.
-For our example, we will use a custom CBC-managed key with the ability to list devices.
-To learn more about the different authentication methods, click `here <https://developer.carbonblack.com/reference/carbon-black-cloud/authentication/>`_.
+identity and authentication, or if you are using VMware Cloud Services Platform, an *OAuth App with Bearer* or
+a *Personal API Token*.  For our example, we will use a custom CBC-managed key with the ability to list devices.
+To learn more about the different authentication methods, click
+`here <https://developer.carbonblack.com/reference/carbon-black-cloud/authentication/>`_.
```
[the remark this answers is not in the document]
abowersox-cb: I'm adding it as `"CustomSDKScript/1.0"`, to emphasize that the `integrationName` property contains a name and version number.

docs/index.rst
```diff
@@ -65,35 +65,9 @@ Get started with Carbon Black Cloud Python SDK here. For detailed information on
...
-   policy
-   recommendations
-   reputation-override
-   unified-binary-store
-   users-grants
-   vulnerabilities
-   watchlists-feeds-reports
-   workload
-
 Full SDK Documentation
```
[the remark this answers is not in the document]
abowersox-cb: I'm just scrapping the extra header and line of text underneath it entirely.  It serves no real purpose.

docs/porting-guide.rst
abowersox-cb: I'm going to have to think about this one, especially since I don't know Observations.

abowersox-cb: Scratch that last. Looks like all I need to do is change up some imports.

docs/alerts.rst
```diff
@@ -203,6 +203,31 @@ The full list of all the attributes can be found in the
...
+
+You can retrieve the event(s) behind a ``CBAnalyticsAlert`` easily with its ``get_events()`` method::
+
+    >>> from cbc_sdk import CBCloudAPI
+    >>> from cbc_sdk.platform import CBAnalyticsAlert
+    >>> api = CBCloudAPI(profile='platform')
+    >>> query = cb.select(CBAnalyticsAlert).set_create_time(range="-4w")
+    >>> # get the first alert returned by the query
+    >>> alert = query[0]
+    >>> for event in alert.get_events():
```
avanbrunt-cb, line 217: How can we make this line standout more as we are trying to call out `get_events` and it gets lost among  the other lines. I recognize this is a pattern we have not applied any where else but does bolding that function make sense? is that enough to make it stand out?

docs/getting-started.rst
```diff
@@ -105,56 +128,39 @@ properties of each retrieved Device object.
...
-  token=ABCDEFGHIJKLMNOPQRSTUVWX/12345678
-  org_key=A1B2C3D4
-  integration_name=MyScript/0.9.0
-
-See the :doc:`authentication` documentation for more information about credentials.
+    >>> for device in devices:
+    ...     print(device.id, device.name, device.last_internal_ip_address, device.last_contact_time)
+    ...
 
 Next Steps
```
avanbrunt-cb, line 163: Can we reference the searching guide specifically as that would be the most logical next step? We should still reference the guides homepage too for more specific use cases

docs/concepts.rst
```diff
@@ -1,595 +0,0 @@
...
-  ...         USB Device ID: {usb.id}
-  ...         USB Device: {usb.vendor_name} {usb.product_name}
-  ...         ''')
-  USB Device ID: 774
-  USB Device: SanDisk Ultra
-
-  USB Device ID: 778
-  USB Device: SanDisk Cruzer Mini
-
-Static Methods
```
avanbrunt-cb, line 517: Do we have a place where we can call out static methods? or is this covered in the specific sub guides

docs/searching.rst
```diff
@@ -0,0 +1,348 @@
...
+
+Text queries may be added to a query object by using the query object's
+:func:`where() <cbc_sdk.base.QueryBuilder.where>`, :func:`and_() <cbc_sdk.base.QueryBuilder.and_>`, and
+:func:`or_() <cbc_sdk.base.QueryBuilder.or_>` methods.  The following example sets up a query looking for events
+in which the program ``googleupdate.exe`` accesses the system registry on a device with a specific hostname, IP
+address, and owner::
+
+    # assume the CBCloudAPI object is in the variable "api"
+    from cbc_sdk.endpoint_standard import Event
+    >>> event_query = api.select(Event).where(hostName='Win10').and_(ipAddress='10.0.0.1')
```
avanbrunt-cb, line 65: It seems odd to use an example with camelCase properties and uses the old integrationService API

docs/searching.rst
```diff
@@ -0,0 +1,348 @@
...
+    >>> event_query.and_(ownerNameExact='DevRel')
+
+The ``where()`` method supplies the initial query parameters, while ``and_()`` and ``or_()`` add additional query
+parameters. As with other languages, ``and_()`` gets grouped together before ``or_()``.
+
+Parameters may either be supplied as text strings or as keyword assignments::
+
+    >>> from cbc_sdk.platform import Device
+    # the following two queries are equivalent
+    >>> string_query = api.select(Device).where("avStatus:AV_ACTIVE")
```
avanbrunt-cb, line 78: avStatus is not a common used property can we use status instead?

docs/searching.rst
```diff
@@ -0,0 +1,348 @@
...
+      "device_os": ["MAC"],
+      "device_os_version": ["10.14.6"]
+    }
+  }
+
+In newer queries, the various specific methods for setting each individual criterion will be replaced with a single
+method::
+
+    # Refine the query with criteria (new style)
+    >>> alert_query.set_criteria("device_os", ["MAC"]).set_criteria("device_os_version", ["10.14.6"])
```
avanbrunt-cb, line 119: It's not `set_criteria` it is `add_criteria`

avanbrunt-cb: https://github.com/carbonblack/carbon-black-cloud-sdk-python/blob/develop/src/cbc_sdk/base.py#L1642

docs/searching.rst
```diff
@@ -0,0 +1,348 @@
...
+
+.. note::
+
+    The ``set_criteria()`` method is supported with Alerts v7, and will be supported in more classes over time.
```
avanbrunt-cb, line 123: We already support this in a bunch of models as `add_criteria` using the CriteriaBuilderSupportMixin

docs/searching.rst
```diff
@@ -0,0 +1,348 @@
...
+    As it becomes supported in existing classes, the existing "specific" methods for setting criteria will be
+    deprecated.
+
+Executing a Query
+-----------------
+
+To execute a query after it's been refined, simply evaluate the query in an *iterable context.*  This may be done
```
avanbrunt-cb, line 130: We should mention the other options such as `first` or `one` from the IterableQueryMixin

docs/searching.rst
```diff
@@ -0,0 +1,348 @@
...
+        }
+    ]
+
+Facet Queries
+*************
+
+More complex facet queries are performed by creating a query *on* a facet type, then refining it as usual, then getting
+the results from the query::
+
+    >>> from cbc_sdk.endpoint_standard import EnrichedEventFacet
```
avanbrunt-cb, line 239: Can we update this to be Observations as EnrichedEvents are deprecated

docs/searching.rst
```diff
@@ -0,0 +1,348 @@
...
+    >>> query = api.select(EnrichedEventFacet).where(process_pid=1000)
+
+Facet queries have two types of special criteria that may be set. One is the ``range`` type which is used to specify
```
avanbrunt-cb, line 242: Range is not faceting specific and should be called out as it's own section this is particularly important for Alerts and Processes

docs/alerts.rst
abowersox-cb: Maybe add an extra comment on the line above? I don't think you can apply bolding to preformatted text.

avanbrunt-cb: Open to options as to what would draw attention to that line in particular. Would be great to define a pattern we can re-use for other examples

docs/getting-started.rst
abowersox-cb: Searching is an important operation in the SDK, and I added a note immediately after this example with the link you want.

docs/concepts.rst
abowersox-cb: This would be best covered in the individual SDK documentation, with support from the guides as necessary.

docs/searching.rst
abowersox-cb: I didn't change the example that was present in the old "concepts" document. If you've got a better idea for this, I'm willing to hear it. Preferably with example code I can just drop in in its place. 😃

docs/searching.rst
abowersox-cb: Again, I didn't write these examples, I just copied them out of "concepts." That said, I'm replacing each one of the "avStatus" values I see in this file with "status."

docs/searching.rst
abowersox-cb: Changing this.

docs/searching.rst
abowersox-cb: Rephrasing.

docs/searching.rst
avanbrunt-cb: You will also need to modify the value being used as the status variable has different values https://developer.carbonblack.com/reference/carbon-black-cloud/platform/latest/devices-api/#base-device

docs/searching.rst
avanbrunt-cb: I'd like you come up with a new example based on what would be most useful to a new user. Alerts or Observations would be to potential options

docs/searching.rst
abowersox-cb: I'm putting a note about those below all the examples of evaluating the query in an iterable context.

docs/searching.rst
abowersox-cb: Do Observations have facets like EnrichedEvents? Because this is an example of a complex facet query.

docs/searching.rst
abowersox-cb: How do you suggest this should be done?

docs/searching.rst
avanbrunt-cb: Yes they should have the same experience as they are both in the investigate API and use the same style of requests

docs/searching.rst
avanbrunt-cb: We should make a section on ranges separate from Facets that explains the set_time_range functions we have in a few queries now

docs/alerts.rst
abowersox-cb: I'm adding the extra comment on the line above, which, as I said, is basically all we _can_ do in a preformatted example.

docs/searching.rst
abowersox-cb: And I am _asking_ for _help,_ because I don't know either of those as well as you do.

docs/searching.rst
abowersox-cb: I changed the "AV_ACTIVE" value to "ACTIVE." That should suffice.

docs/alerts.rst
avanbrunt-cb: What do you think on how well it actually makes it stand out? or does it just add more text?

docs/searching.rst
avanbrunt-cb: Make an alert search request with a `parent_name` and `process_name`. Look for an interesting alert in the prod02 partner org and take it's parent_name and process_name. You could even use the upcoming new `Alert` model name too

docs/alerts.rst
kebringer-cb: Good enough for now.  The Alerts guide will have significant updates with the v7 changes in SDK 1.5.0.  Likely will become getting observations for an Alert instead of events.

docs/searching.rst
abowersox-cb: I am replacing this with an Observation query, because if I try to run an Alert query, it'll be with Alerts v6 and I think you might object if it weren't with Alerts v7. Observation is safer. And this is just an example of query-building syntax anyway.

docs/searching.rst
abowersox-cb: I may have something for this.

docs/searching.rst
abowersox-cb: I've got an ObservationFacet example.

docs/getting-started.rst
```diff
@@ -105,56 +128,42 @@ properties of each retrieved Device object.
...
+    >>> from cbc_sdk.platform import Device
+    >>> cb = CBCloudAPI(profile='default')
 
 This imports the necessary classes and creates an instance of the base ``CBCloudAPI`` object.  By default, the file
 credentials provider is used. We set it to use the ``default`` profile in your ``credentials.cbc`` file, which you
 set up earlier.
 
-**N.B.:** On Windows, a security warning message will be generated about file access to CBC SDK credentials being
+.. note:: On Windows, a security warning message will be generated about file access to CBC SDK credentials being
 inherently insecure.
```
kebringer-cb, line 145: This line break pushes the text outside the blue note box.

Screenshot 2023-09-07 at 7 24 49 AM

avanbrunt-cb (approved):

docs/getting-started.rst
abowersox-cb: Fixed.

kebringer-cb (approved):

PR #395, opened 2023-09-06
Cbapi 4915 json mapping bugs

heiderje-vmware:
## Pull request checklist

- [ x] New and existing tests pass locally with the changes.

[CBAPI-4915](https://jira.carbonblack.local/browse/CBAPI-4915)

- [ ] Yes
- [ x] No


avanbrunt-cb (approved):

src/cbc_sdk/platform/alerts.py
```diff
@@ -84,47 +84,48 @@ class Alert(PlatformModel):
     }
 
     REMAPPED_ALERTS_V7_TO_V6 = {
```
kebringer-cb, line 86: In the to_json() method in the "if version == "v6" section where this is called around line 470, the "if key ==threat_cause_event_id" check is checking for the v6 field name when it is iterating over the v7 fields.

Instead of:
```
if key == "threat_cause_event_id":
                    modified_json["created_by_event_id"] = value
```
 
I think it should be:
              ```
  if key == "primary_event_id": # the v7 field name
                    # created_by_event_id set by REMAPPED_ALERTS_V7_TO_V6
                    # secondary mapping when alert type is CB_ANALYTICS needed to created_by_event_id
                    modified_json["threat_cause_cause_event_id"] = value # the other v6 field name that also holds the primary event id value
```

Comment (kebringer-cb) [inline src/cbc_sdk/platform/alerts.py:86]: Two mappings are needed for process_guid

* Exists on all alert types in v7
* Must be mapped to process_guid on WATCHLIST type alerts in v6
* must also be mapped to threat_cause_process_guid when the alert type is CB_ANALYTIC

File: src/cbc_sdk/platform/alerts.py [hunk 2]
```diff
@@ -84,47 +84,48 @@ class Alert(PlatformModel):
...
-        "process_sha256": "threat_cause_actor_sha256",
-        "primary_event_id": "threat_cause_cause_event_id",
-        "process_md5": "threat_cause_md5",
-        "parent_guid": "threat_cause_parent_guid",
-        "process_reputation": "threat_cause_reputation",
         "ttps": "threat_indicators",
         "watchlists.id": "watchlists",
         "workflow.change_timestamp": "workflow.last_update_time",
-        "workflow.note": "workflow.comment",
         "workflow.closure_reason": "workflow.remediation",
```
kebringer-cb, line 126: Remove mapping 
"workflow.closure_reason": "workflow.remediation",
Received new info - remediation is free text, closure_reason is enum, they do not have the same value.

kebringer-cb: Further work will be required for workflow, tracked on CBAPI-4807.
For this PR, remove the mapping between closure_reason and remediation.

kebringer-cb (approved):

PR #403, opened 2023-09-20
Feature cis benchmark

hkaragitliev-cb:
## Pull request checklist

## Pull Request Description

Some code snippets:

`
### Get Organization Settings
```
ComplianceBenchmark.get_compliance_schedule(cb)
```

### Set Organization Settings
```
ComplianceBenchmark.set_compliance_schedule(cb, scan_schedule="RRULE:FREQ=DAILY;BYHOUR=17;BYMINUTE=30;BYSECOND=0", scan_timezone="UTC")
```

### Search Benchmark Sets
```
benchmark_sets = cb.select(ComplianceBenchmark)
print(*benchmark_sets)
```

### Search Rules
```
benchmark_sets = cb.select(ComplianceBenchmark)
rules = list(benchmark_sets[0].search_rules())
print(*rules)
```

### Get Specified Rule
```
benchmark_sets = cb.select(ComplianceBenchmark)
rule = list(benchmark_sets[0].search_rules('00869D86-6E61-4D7D-A0A3-6F5CDE2E5753'))
print(*rule)
```

### Get All Benchmark Set Sections
```
benchmark_sets = cb.select(ComplianceBenchmark)
set_sections = list(benchmark_sets[0].get_set_sections())
print(*set_sections)
```

### On a Benchmark Set
```
benchmark_sets = cb.select(ComplianceBenchmark)
benchmark_sets[0].execute_action(action='reassess')
```

`

- [ ] Yes
- [ ] No


src/cbc_sdk/workload/compliance_assessment.py
```diff
@@ -0,0 +1,465 @@
...
+
+        self._query_builder = QueryBuilder()
+        self._criteria = {}
+        self._sortcriteria = {}
+        self._total_results = 0
+
+        self._benchmark_set_id = None
+        self._rule_id = None
+
+    def add_criteria(self, key, value, operator='EQUALS'):
```
avanbrunt-cb, line 300: I don't think CIS benchmarks uses operator so we can use the functions from https://github.com/carbonblack/carbon-black-cloud-sdk-python/blob/90794cdfdfc69dfc4f6f87989e89da3b3522caf7/src/cbc_sdk/base.py#L1638

src/cbc_sdk/workload/compliance_assessment.py
```diff
@@ -0,0 +1,465 @@
...
+            >>> print(*rules)
+            >>> # To return a single rule document, add the rule ID.
+            >>> rule = list(benchmark_sets.search_rules('00869D86-6E61-4D7D-A0A3-6F5CDE2E5753'))
+            >>> print(rule)
+        """
+        if rule_id is not None:
+            self._rule_id = rule_id
+            url = self.urlobject.format(self._cb.credentials.org_key) + f"{self.id}/rules/{rule_id}"
+            result = self._cb.get_object(url)
+            yield ComplianceBenchmark(self._cb, initial_data=result)
```
avanbrunt-cb, line 166: It seems odd to return ComplianceBenchmark as that is a different data model. We should create a sub class for ComplainceBenchmarkRule instead ComplianceBenchmark as its possible to update the rule to be Enabled/Disabled.

src/cbc_sdk/workload/compliance_assessment.py
```diff
@@ -0,0 +1,465 @@
...
+        if not scan_schedule or not scan_timezone or scan_schedule == "" or scan_timezone == "":
+            raise ApiError(
+                "scan_schedule and scan_timezone are required. http://developer.carbonblack.com/reference/carbon-black-cloud/cb-liveops/latest/livequery-api/#recurrence-rules")
+
+        args = {"scan_schedule": scan_schedule, "scan_timezone": scan_timezone}
+        url = f"/compliance/assessment/api/v1/orgs/{cb.credentials.org_key}/settings"
+
+        return cb.put_object(url, body=args).json()
+
+    def search_rules(self, rule_id=None):
```
avanbrunt-cb, line 142: We don't provide a mechanism to provide a query or criteria

src/cbc_sdk/workload/compliance_assessment.py
```diff
@@ -0,0 +1,465 @@
...
+                yield ComplianceBenchmark(self._cb, initial_data=item)
+
+            current += len(results)
+            if max_rows > 0 and current >= max_rows:
+                break
+
+            if current >= self._total_results:
+                break
+
+    def get_set_sections(self):
```
avanbrunt-cb, line 191: I would rename this to `get_sections` as the model is already the BenchmarkSet

src/cbc_sdk/workload/compliance_assessment.py
```diff
@@ -0,0 +1,465 @@
...
+        Example:
+            >>> cb = CBCloudAPI(profile="example_profile")
+            >>> benchmark = cb.select(ComplianceBenchmark).set_benchmark_set_id('benchmark123')
+            >>> for section in benchmark.get_benchmark_set_sections():
+            ...     print(section.section_name, section.section_id)
+        """
+        url = self.urlobject.format(self._cb.credentials.org_key) + f"{self.id}/sections"
+        results = self._cb.get_object(url)
+        for item in results:
+            yield ComplianceBenchmark(self._cb, initial_data=item)
```
avanbrunt-cb, line 208: We could return the raw json list no need to return the ComplianceBenchmark

src/cbc_sdk/workload/compliance_assessment.py
```diff
@@ -0,0 +1,465 @@
...
+        Execute a specified action on devices within on a Benchmark Set, or specific devives
+        within a Benchmark Set only.
+
+        Args:
+            action (str): The action to be executed. Available:
+                'ENABLE': Enable the object.
+                'DISABLE': Disable the object.
+                'REASSESS': Reassess the object.
+
+            device_ids (str or list, optional): IDs of devices on which the action will be executed.
```
avanbrunt-cb, line 221: Don't the ids need sent as integers not strings

src/cbc_sdk/workload/compliance_assessment.py
```diff
@@ -0,0 +1,465 @@
...
+            raise ApiError(message)
+
+        args = {"action": action}
+        if device_ids:
+            args['device_ids'] = [device_ids] if isinstance(device_ids, str) else device_ids
+
+        url = self.urlobject.format(self._cb.credentials.org_key) + f"{self.id}/actions"
+        return self._cb.post_object(url, body=args).json()
+
+    def get_benchmark_set_summary(self):
```
avanbrunt-cb, line 254: Rename to get_summary as this is already the BenchmarkSet

src/cbc_sdk/workload/compliance_assessment.py
```diff
@@ -0,0 +1,465 @@
...
+        fetches the summary data using the Carbon Black API, and yields Compliance objects for each item in the summary.
+
+        Returns:
+            generator of ComplianceBenchmark: A generator yielding Compliance objects, each representing a summary item.
+        """
+        url = self.urlobject.format(self._cb.credentials.org_key) + f"{self.id}/compliance/summary"
+        results = self._cb.get_object(url)
+        for item in results:
+            yield ComplianceBenchmark(self._cb, initial_data=item)
+
```
avanbrunt-cb, line 268: We need to look into adding the following endpoints but I think we can enable them as single functions inside ComplianceBenchmark as long as we expose the query, sort and criteria and return the raw json response
- [Search Device Compliance Summaries](https://developer.carbonblack.com/reference/carbon-black-cloud/workload-protection/latest/cis-benchmark-api/#search-device-compliance-summaries)
- [Search Rule Compliance Summaries](https://developer.carbonblack.com/reference/carbon-black-cloud/workload-protection/latest/cis-benchmark-api/#search-rule-compliance-summaries)
- [Search Rule Compliance Results for a Device](https://developer.carbonblack.com/reference/carbon-black-cloud/workload-protection/latest/cis-benchmark-api/#search-rule-compliance-results-for-a-device)
- [Search Rule Results for Devices](https://developer.carbonblack.com/reference/carbon-black-cloud/workload-protection/latest/cis-benchmark-api/#search-rule-results-for-devices)

src/cbc_sdk/workload/compliance_assessment.py
hkaragitliev-cb: Yes, but I convert them to integers anyway, rather than throw an error.

avanbrunt-cb: The problem is we would source them as ints already then expect to convert them to strings then back again. I'd recommend using int still instead of str

docs/compliance.rst
```diff
@@ -0,0 +1,88 @@
...
+                    type: Custom
+            update_time: 2024-04-15T21:24:43.283032Z
+            updated_by:
+                version: 1.0.0.4
+
+
+Modify Compliance Benchmarks Schedule
+---------------
+
+By using the following the example, you can get and set the benchmark assessment schedule
```
emitreva-cb, line 58: Maybe an example of set would be nice.

src/cbc_sdk/workload/compliance_assessment.py
```diff
@@ -0,0 +1,599 @@
...
+        Args:
+            cb (CBCloudAPI): Instance of CBCloudAPI.
+
+        Returns:
+            ComplianceBenchmarkQuery: Query implementation for ComplianceBenchmark.
+        """
+        return ComplianceBenchmarkQuery(cls, cb)
+
+    def get_benchmark_set_summary(self):
```
emitreva-cb, line 66: Missing permissions everywhere.

src/cbc_sdk/workload/compliance_assessment.py
```diff
@@ -0,0 +1,599 @@
+#!/usr/bin/env python3
+
+# *******************************************************
+# Copyright (c) VMware, Inc. 2021-2023. All Rights Reserved.
```
emitreva-cb, line 4: Not that important, but maybe an update of the year.

src/cbc_sdk/workload/compliance_assessment.py
```diff
@@ -0,0 +1,599 @@
...
+
+        Example:
+            >>> cb = CBCloudAPI(profile="example_profile")
+            >>> benchmark_set = cb.select(ComplianceBenchmark).first()
+            >>> rules = benchmark_set.get_rule_compliance()
+        """
+        url = self.urlobject.format(self._cb.credentials.org_key) + f"{self.id}/compliance/rules/_search"
+        current = 0
+        rule_compliances = []
+        while True:
```
emitreva-cb, line 322: Seems like most of the code for getting the rules for specific device or benchmark set is the same, can we extract it to a non-public function to reuse?

docs/compliance.rst
avanbrunt-cb: The example includes setting

src/cbc_sdk/workload/compliance_assessment.py
avanbrunt-cb: They are very similar but the sorts are different too so I'd rather keep them separate given the code is simple and in case we want/need to adjust the functionality further in the future

kebringer-cb (commented):
Approved pending a decision on whether GET compliance summary API is going to be fixed on the backend or removed from the SDK.

kebringer-cb (approved):

abowersox-cb (approved):

PR #414, opened 2023-09-28
CBAPI-4946: Apply style guide to Devices module

abowersox-cb:
## Pull request checklist

https://jira.carbonblack.local/browse/CBAPI-4946

## Pull Request Description

Changed docstrings for platform devices module to bring them into line with the Docstring Style Guide.

- [ ] Yes
- [x] No

## How Has This Been Tested?

Changes have been previewed by generating Sphinx docs and eyeballing them in a browser.


src/cbc_sdk/platform/devices.py
```diff
@@ -298,14 +317,19 @@ def nsx_remediation(self, tag, set_tag=True):
 
 
 class DeviceFacet(UnrefreshableModel):
-    """Represents a device field in a facet search."""
+    """
+    Represents a device field in a facet search.
+
+    *Faceting* is a search technique that categorizes search results according to common attributes. This allows
```
avanbrunt-cb, line 323: We did an example on the Device we should add one for DeviceFacet

src/cbc_sdk/platform/devices.py
```diff
@@ -406,7 +430,12 @@ def values_(self):
 
 class DeviceSearchQuery(BaseQuery, QueryBuilderSupportMixin, CriteriaBuilderSupportMixin,
                         IterableQueryMixin, AsyncQueryMixin):
-    """Represents a query that is used to locate Device objects."""
+    """
+    Query object that is used to locate ``Device`` objects.
+
+    The ``DeviceSearchQuery`` is constructed via SDK functions like the ``select()`` method on ``CBCloudAPI``.
+    The user would then add a query and criteria to it before iterating over the results.
```
avanbrunt-cb, line 437: This should be and/or not just and

avanbrunt-cb (approved):
Two small comments otherwise this looks good. Will need flake8 error fixed as well

src/cbc_sdk/platform/devices.py
abowersox-cb: Doing this.

src/cbc_sdk/platform/devices.py
avanbrunt-cb: I do not you'll have to try it and create one

abowersox-cb: I have one, which is different from the one on the `device_query()` method.

src/cbc_sdk/platform/devices.py
```diff
@@ -298,14 +317,25 @@ def nsx_remediation(self, tag, set_tag=True):
...
+    """
+    Represents a device field in a facet search.
+
+    *Faceting* is a search technique that categorizes search results according to common attributes. This allows
+    users to explore and discover information within a dataset, in this case, the set of devices.
+
+    Example:
+        >>> facets = api.select(Device).facets(['policy_id'])
+        >>> for value in facets[0].values_:
+        ...     count = len(value.query_devices())
```
avanbrunt-cb, line 329: This seems excessive to make another query to get the count for each value as it has the total count in the response which should be stored in DeviceFacetValue as `total`

https://developer.carbonblack.com/reference/carbon-black-cloud/platform/latest/devices-api/#facet-devices

abowersox-cb: That worked, so I'm doing it that way.

avanbrunt-cb:
Looks good to me!

src/cbc_sdk/platform/devices.py
```diff
@@ -370,9 +405,6 @@ def _query_implementation(cls, cb, **kwargs):
         Args:
             cb (BaseAPI): Reference to API object used to communicate with the server.
             **kwargs (dict): Not used, retained for compatibility.
-
```
emitreva-cb, line 373: Should we have the Raises section here or it is useless considering it was stated that it is not used?

src/cbc_sdk/platform/devices.py
```diff
@@ -460,9 +492,6 @@ def set_ad_group_ids(self, ad_group_ids):
 
         Returns:
             DeviceSearchQuery: This instance.
-
-        Raises:
```
emitreva-cb, line 464: Why is the Raises section removed for this one and the next few ones?

src/cbc_sdk/platform/devices.py
abowersox-cb: From the Docstring Style Guide: "Exceptions that get raised if the API in the docstring is violated **SHOULD NOT** be documented here, because this would mean that behavior under violation of the API would be _part_ of the API."

emitreva-cb (approved):

PR #424, opened 2023-10-13
Alert notes

avanbrunt-cb:
## Pull request checklist

CBAPI-4766

## Pull Request Description

Add threat notes and alert/threat history

- [ ] Yes
- [x] No

## How Has This Been Tested?

Unit tests and manual E2E

### DIFF
--- a/src/cbc_sdk/platform/alerts.py
+++ b/src/cbc_sdk/platform/alerts.py
@@ -202,21 +202,19 @@ class Alert(PlatformModel):
         """Represents a note within an alert."""
         REMAPPED_NOTES_V6_TO_V7 = {
             "create_time": "create_timestamp",
-
         }
 
         REMAPPED_NOTES_V7_TO_V6 = {
             "create_timestamp": "create_time",
-
         }
 
         urlobject = "/api/alerts/v7/orgs/{0}/alerts/{1}/notes"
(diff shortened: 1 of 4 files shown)


src/cbc_sdk/platform/alerts.py
```diff
@@ -198,36 +199,53 @@ def get_observations(self, timeout=0):
         obs = Observation.bulk_get_details(self._cb, alert_id=alert_id, timeout=timeout)
         return obs
 
+    def get_history(self, threat=False):
+        """
+        Get the actions taken on an Alert such as Notes added and workflow state changes.
+
```
kebringer-cb, line 205: needs markup for param of threat

src/cbc_sdk/platform/alerts.py
```diff
@@ -198,36 +199,53 @@ def get_observations(self, timeout=0):
...
+        Returns:
```
emitreva-cb, line 206: Based on Amy's guidelines this should not have a Returns section considering that other methods in this file explain what they are returning in the description of the methods.

emitreva-cb (approved):

src/cbc_sdk/platform/alerts.py
avanbrunt-cb: I disagree as we need to add returns in some cases like these to clarify what is specifically being returned. I think returns can be optional but it's not required to be dropped

kebringer-cb (approved):

PR #448, opened 2023-11-27
Cbapi 5087 searched group alert helper function

heiderje-vmware:
## Pull request checklist

- [x ] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [x ] Tests have been added that prove the fix is effective or that the feature works.
- [x ] New and existing tests pass locally with the changes.

[CBAPI-5087](https://jira.carbonblack.local/browse/CBAPI-5087)

## Pull Request Description
adds a helper function to the GroupAlert object that creates the AlertSearchQuery for all alerts associated with a given GroupAlert object

- [ ] Yes
- [ x] No

## How Has This Been Tested?
added to common scenarios

### DIFF
--- a/src/cbc_sdk/platform/alerts.py
+++ b/src/cbc_sdk/platform/alerts.py
@@ -980,7 +980,7 @@ class GroupedAlert(PlatformModel):
         """
         return self._most_recent_alert
 
-    def get_alerts(self):
+    def get_alert_search_query(self):
         """
         Returns the Alert Search Query needed to pull all alerts for a given Group Alert.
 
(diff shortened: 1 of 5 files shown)


src/cbc_sdk/platform/alerts.py
```diff
@@ -979,6 +980,25 @@ def most_recent_alert_(self):
         """
         return self._most_recent_alert
 
+    def get_alerts(self):
```
kebringer-cb, line 983: Since this returns the alert query object rather than a list of alerts, I think either
* the method name should be changed to get_alerts_query
or
* change the method should return a list.  `return alert_search_query.all()` will achieve this.

Agree it works as designed, but isn't intuitive to use.

src/cbc_sdk/platform/alerts.py
```diff
@@ -979,6 +980,34 @@ def most_recent_alert_(self):
...
+
+    def get_alerts(self):
+        """
+        Returns the all alerts for a given Group Alert.
+
+        Returns:
+            list: alerts associated with the calling group alert.
+        """
+        return self.get_alert_search_query().all()
+
```
abowersox-cb, line 1010: This is as discussed. Well done.

PR #450, opened 2023-11-30
CBAPI-5151: Add devices pagination

mrangelov-cb:
## Pull request checklist

- [ ] Yes
- [x] No


src/cbc_sdk/platform/devices.py
```diff
@@ -468,6 +468,23 @@ def __init__(self, doc_class, cb):
         self._exclusions = {}
         self._sortcriteria = {}
         self.max_rows = -1
+        self.start = -1
+        self.num_found = 0
+
+    def set_start(self, start):
```
avanbrunt-cb, line 474: This is not needed as we can get the results using `query[5:10`

avanbrunt-cb (changes requested):
Can you summarize why the current pagination support we have doesn't work and what problem this is trying to solve. This is introducing a new pagination style that we should uniformly introduce too more models assuming there is a use case we can't currently achieve today

mrangelov-cb:
The current pagination would work, yet in our particular case it is inefficient. The reason for that is that the request to CBC is made from within a request handler in the Flask app and we cannot maintain context (CBCloudAPI and DeviceQuery) across subsequent HTTP requests to that route. On the UI side it is relatively easy to maintain state and keep track which rows we are currently visualising, but, whenever a request to the Flask backend is made (e.g "get me the list of devices from 300th to 400th"), in views.py we need to reinstantiate CBCloudAPI and do the device search again and since we cannot provide a starting row, that would always be from row 0. We could in theory maintain a pool of DeviceQuery objects on the Flask side if we had the session cookie, but the request to Flask is made via the QRadar app proxy, not directly by the browser and session information is stripped (we can only get the QRadar username doing the request, but that is not enough).

mrangelov-cb:
This would work, my concern is that it would be much slower than searching devices with search offset as we need to get up to 10000 devices while we only need the "page size" number of them and UI-wise it could be unacceptably slow to browse the list of devices.

avanbrunt-cb:
Should this be a separate API call outside of CBC SDK device model selecting?

mrangelov-cb:
Hello,

I managed to get access to a work with 10000 endpoints (not real) provided by QA.

I did a quick test with two quick python scripts:

```
mrangelov@mrangelovYMD6R /tmp % cat a.py 
from cbc_sdk.platform import Device

cb = CBCloudAPI()
devs = cb.select(Device).set_max_rows(10000)
last_page = devs[9900:10000]
data = []
for dev in last_page:
    data.append(dev._info)

mrangelov@mrangelovYMD6R /tmp % cat b.py 
from cbc_sdk.platform import Device

cb = CBCloudAPI()
devs = cb.select(Device).set_max_rows(100).set_start(9900)
data = []
for dev in devs:
    data.append(dev._info)
```

Those are the timings:

```
mrangelov@mrangelovYMD6R /tmp % time python3 a.py 
python3 a.py  1.27s user 0.20s system 29% cpu 5.015 total
mrangelov@mrangelovYMD6R /tmp % time python3 b.py
python3 b.py  0.70s user 0.12s system 49% cpu 1.644 total
```

So the difference is ~4.4 seconds.

I slightly changed the scripts to get devices from 1000th to 1100th and the results are similar (which is a bit surprising):

```
mrangelov@mrangelovYMD6R /tmp % cat a.py 
from cbc_sdk.platform import Device

cb = CBCloudAPI()
devs = cb.select(Device).set_max_rows(10000)
last_page = devs[900:1000]
data = []
for dev in last_page:
    data.append(dev._info)
                                                                                                                                                          
mrangelov@mrangelovYMD6R /tmp % cat b.py 
from cbc_sdk.platform import Device

cb = CBCloudAPI()
devs = cb.select(Device).set_max_rows(100).set_start(900)
data = []
for dev in devs:
    data.append(dev._info)
mrangelov@mrangelovYMD6R /tmp % time python3 a.py 
python3 a.py  1.25s user 0.19s system 25% cpu 5.583 total
mrangelov@mrangelovYMD6R /tmp % time python3 b.py
python3 b.py  0.72s user 0.12s system 51% cpu 1.631 total
```

However in a.py when I modify set_max_rows from 10000 to 1000 it makes sense:

```
time python3 a.py
python3 a.py  0.77s user 0.13s system 40% cpu 2.243 total
```


Looks like I made a mistake with the limitations of paging in the Devices API though, as I reread the developer network (need to address that in the code). There is no start+rows limit of 10000, instead we have rows limit of 10000 and start limit of 200000 (how do orgs with more than 200k endpoints work?)

Anyway, this leads me to the conclusion that the workaround (without providing start) gets slower and slower as the number of devices grows. With the workaround, we have a hard limit of rows up to 10k (would that still work if we need to get the 10001th device?). With the introduction of start parameter we can fetch up to 200k+10k devices and the execution time remains the same (around 1.6 seconds). With the workaround, the execution time increases with the overall number of devices that we have in the org.

mrangelov-cb:
Actually my tests were wrong, I was doing them on my branch. I switched back to develop and rerun time python3 a.py and now I get similar timing. So my only question would be, would it be possible (using just set_max_rows and slicing) to get devices past the 10000 set_max_rows limit?

avanbrunt-cb:
> Actually my tests were wrong, I was doing them on my branch. I switched back to develop and rerun time python3 a.py and now I get similar timing. So my only question would be, would it be possible (using just set_max_rows and slicing) to get devices past the 10000 set_max_rows limit?

As long as we didn't add any SDK checks we should be able to continue to paginate. Additionally if we want to get devices beyond 200k that is the intent behind the `_scroll` endpoint though that requires maintaining state with the `search_after` value.

Something we should consider is what problem are we trying to solve in QRadar with the device tab

mrangelov-cb:
The idea was to solve an issue that we had with orgs with large number of devices as they were all fetched and passed to the UI to do the pagination in the frontend. Once we started working on that, the next potential issue that we wanted to avoid was having an unresponsible UI as we are going to make a search request whenever we change the page and having to wait a long time wasn't desired. However, the timing tests convince me that no start parameter is needed and performance does not degrade linearly with the number of devices in the org. I will do some more tests to ensure we can get past max_rows by getting slices past it and if it works fine, I will drop that PR as it would not be needed.

avanbrunt-cb:
That makes sense. I was also trying to frame why do we even have a device tab what problem are we solving for customers by having a device view in the app. By focusing on the problem we are trying to solve we can account for performance issues in other ways too. Do we really want to paginate 500k devices or should we be focusing more on device details and pivots back to the CBC console? or is this designed to given non CBC console users access.

Though based on the performance numbers I would say lets stay away from adding a `set_start` and if we truly need more granular fetching then we should approach that with a different mechanism that iterating the query

avanbrunt-cb:
Not sure I see what issue you are referencing

mrangelov-cb:
devs[0:5] gets devices from index 0 to 4 as expected
devs[5:10] gets devices from index 4 to 8 (not 5..9)
devs[0:10] gets devices from 0 to 9 properly

This is a little bit unexpected, e.g:

```
>>> result = [0,1,2,3,4,5,6,7,8,9]
>>> result[0:5]
[0, 1, 2, 3, 4]
>>> result[5:10]
[5, 6, 7, 8, 9]
>>> result[0:10]
[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
```

PR #451, opened 2023-12-01
[PENDING] Grouped alert build request

kebringer-cb:
## Pull request checklist
TESTS WILL NOT PASS UNTIL CBAPI-5088 (and maybe CBAPI-5087) are merged to the feature branch and available for this PR.

## Pull Request Description

Added  and close() method to GroupedAlertSearchQuery to override the version inherited from AlertSearchQuery.

_build_request needed to include get group_by field.
This then caused the close() function to incorrectly have group_by, so the close() method also needed to be overridden.

- [ ] Yes
- [ ] No

### DIFF
--- a/src/cbc_sdk/platform/alerts.py
+++ b/src/cbc_sdk/platform/alerts.py
@@ -1586,7 +1586,8 @@ class GroupedAlertSearchQuery(AlertSearchQuery):
         Returns:
             dict: The complete request body.
         """
-        request = super()._build_request(from_row, max_rows, add_sort=True)
+        # request = super()._build_request(from_row, max_rows, add_sort=True)
+        request = super(GroupedAlertSearchQuery, self)._build_request(from_row, max_rows, add_sort=True)
         request["group_by"] = {"field": self._group_by}
         return request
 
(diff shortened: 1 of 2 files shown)


src/cbc_sdk/platform/alerts.py
```diff
@@ -1574,6 +1574,22 @@ def set_group_by(self, field):
         self._group_by = field
         return self
 
+    def _build_request(self, from_row, max_rows, add_sort=True):
```
avanbrunt-cb, line 1577: We already do this in `_perform_query` so if we are going to define this function we need to update the following

```
while still_querying:
    request = self._build_request(current, max_rows)
    request["group_by"] = {"field": self._group_by}
```

src/cbc_sdk/platform/alerts.py
```diff
@@ -1574,6 +1574,22 @@ def set_group_by(self, field):
...
+
+        Args:
+            from_row (int): The row to start the query at.
+            max_rows (int): The maximum number of rows to be returned.
+            add_sort (bool): If True(default), the sort criteria will be added as part of the request.
+
+        Returns:
+            dict: The complete request body.
+        """
+        request = super()._build_request(from_row, max_rows, add_sort=True)
```
avanbrunt-cb, line 1589: This needs to be `super(GroupedAlertSearchQuery, self)`

src/cbc_sdk/platform/alerts.py
```diff
@@ -1618,3 +1634,30 @@ def _perform_query(self, from_row=1, max_rows=-1):
             if current >= self._total_results:
                 still_querying = False
                 break
+
+    def close(self, closure_reason=None, determination=None, note=None, ):
```
avanbrunt-cb, line 1638: I'm confused why we want to implement this in GroupedAlertSearchQuery as it only covers a single workflow update and affects all alerts not just the alert of a single grouped alert. It seems like we'd rather have users call the code themselves
```
alert_query = self.get_alert_search_query()
alert_query.close(closure_reason, note, determination)
```

src/cbc_sdk/platform/alerts.py
kebringer-cb: removed from _perform_query

src/cbc_sdk/platform/alerts.py
kebringer-cb: That's fair. Have changed it to raise a NotImplemented exception.
It needs something to override the Alert implementation.

kebringer-cb:
Replaced by https://github.com/carbonblack/carbon-black-cloud-sdk-python/pull/453

PR #452, opened 2023-12-04
New methods to convert an alertSearchQuery to a groupedAlertSearchQue…

heiderje-vmware:
## Pull request checklist

- [ x] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [ x] Tests have been added that prove the fix is effective or that the feature works.
- [ x] New and existing tests pass locally with the changes.

[CBAPI_5088](https://jira.carbonblack.local/browse/CBAPI-5088)

## Pull Request Description
adds methods to convert from alertSearchQuery to GroupAlertSearchQuery and vice versa 

- [ ] Yes
- [ x] No

## How Has This Been Tested?
example code and unit tests

### DIFF
--- a/src/cbc_sdk/platform/alerts.py
+++ b/src/cbc_sdk/platform/alerts.py
@@ -34,6 +34,7 @@ from backports._datetime_fromisoformat import datetime_fromisoformat
 """Alert Models"""
 
 MAX_RESULTS_LIMIT = 10000
+REQUEST_IGNORED_KEYS = ["_doc_class", "_cb", "_count_valid", "_total_results", "_query_builder", "_sortcriteria"]
 
 
 class Alert(PlatformModel):
@@ -990,14 +991,10 @@ class GroupedAlert(PlatformModel):
         Note:
             Does not preserve sort criterion
         """
-        ignored_keys = ["_doc_class", "_cb", "_count_valid", "_total_results", "_sortcriteria", "_query_builder"]
         alert_search_query = self._cb.select(Alert)
         for key, value in vars(alert_search_query).items():
-            if hasattr(self._request, key) and key not in ignored_keys:
+            if hasattr(self._request, key) and key not in REQUEST_IGNORED_KEYS:
                 setattr(alert_search_query, key, self._request.__getattribute__(key))
-        key = "_time_range"
-        if hasattr(self._request, key):
-            setattr(alert_search_query, key, self._request.__getattribute__(key))
 
         alert_search_query.add_criteria(self._request._group_by.lower(), self.most_recent_alert["threat_id"])
         return alert_search_query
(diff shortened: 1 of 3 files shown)


kebringer-cb (approved):

src/cbc_sdk/platform/__init__.py
```diff
@@ -4,7 +4,7 @@
 
 from cbc_sdk.platform.alerts import (Alert, WatchlistAlert, CBAnalyticsAlert, DeviceControlAlert,
                                      ContainerRuntimeAlert, HostBasedFirewallAlert, IntrusionDetectionSystemAlert,
-                                     GroupedAlert)
+                                     GroupedAlert, AlertSearchQuery, GroupedAlertSearchQuery)
```
avanbrunt-cb, line 7: Why are we adding this to the `__init__`? We don't make Query classes in the package level

src/cbc_sdk/platform/alerts.py
```diff
@@ -1603,6 +1631,25 @@ def set_group_by(self, field):
...
+        Note: Does not preserve sort criterion
+        """
+        ignored_keys = ["_doc_class", "_cb", "_count_valid", "_total_results", "_query_builder", "_sortcriteria"]
+        alert_search_query = self._cb.select(Alert)
+        for key, value in vars(alert_search_query).items():
+            if hasattr(self, key) and key not in ignored_keys:
+                setattr(alert_search_query, key, self.__getattribute__(key))
+        key = "_time_range"
+        if hasattr(self, key):
+            setattr(alert_search_query, key, self.__getattribute__(key))
```
avanbrunt-cb, line 1649: I'd avoid making a temporary variable here as it leads to a bit of confusion with the for loop above. Also is there a reason the for loop would not catch this it seems redundant

avanbrunt-cb: Same applies for set_group_by

avanbrunt-cb (changes requested):
A few outstanding comments

src/cbc_sdk/platform/__init__.py
heiderje-vmware: I was trying to access the query classes for my unit tests, to assert the return types on conversion tests. If acceptable i can do it by __module__ and __name__

avanbrunt-cb: Yeah for unit testing I would use `from cbc_sdk.platform.alerts import AlertSearchQuery, GroupedAlertSearchQuery` instead of adding it to the `__init__`

src/cbc_sdk/platform/alerts.py
avanbrunt-cb: As discussed in the meeting, we should set a default `self._time_range = {}` in order to reduce unique logic for individual properties

avanbrunt-cb (approved):

PR #455, opened 2023-12-06
Updated build_request to set group_by and overrode close

kebringer-cb:
## Pull request checklist

- [X] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [X] Tests have been added that prove the fix is effective or that the feature works.
- [X] New and existing tests pass locally with the changes.
- [X] Code follows the style guidelines of this project (PEP8, clean code, linter).
- [X] A self-review of the code has been done.

## Pull Request Description

* _build_request for GroupedAlertSearchQuery needed to be overridden to set the group_by field.  Without this getting the length of the GroupedAlertResultSet fails.
* the Close method needed to be overridden from AlertSearchQuery to raise a not implemented exception.

- [ ] Yes
- [X] No

## How Has This Been Tested?

Run locally and added unit tests

### DIFF
--- a/src/cbc_sdk/platform/alerts.py
+++ b/src/cbc_sdk/platform/alerts.py
@@ -1603,6 +1603,22 @@ class GroupedAlertSearchQuery(AlertSearchQuery):
         self._group_by = field
         return self
 
+    def _build_request(self, from_row, max_rows, add_sort=True):
+        """
+        Creates the request body for an API call.
+
+        Args:
+            from_row (int): The row to start the query at.
+            max_rows (int): The maximum number of rows to be returned.
+            add_sort (bool): If True(default), the sort criteria will be added as part of the request.
+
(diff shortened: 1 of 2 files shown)


abowersox-cb (approved):

src/cbc_sdk/platform/alerts.py
```diff
@@ -1648,3 +1663,17 @@ def _perform_query(self, from_row=1, max_rows=-1):
             if current >= self._total_results:
                 still_querying = False
                 break
+
+    def close(self, closure_reason=None, determination=None, note=None, ):
```
avanbrunt-cb, line 1667: I'm still confused is reason we are doing this to prevent the inherited methods from working as the other workflow ones are inherited too

avanbrunt-cb (approved):

PR #461, opened 2023-12-14
CPAPI-4255: Asset Group Support in CBC SDK p3 - Member Functions

abowersox-cb:
## Pull request checklist

https://jira.carbonblack.local/browse/CBAPI-4255

## Pull Request Description

Added functions to `AssetGroup` to list group members, add group members, and remove group members, and functions to `Device` to list the asset groups of which it's a member.

- [ ] Yes
- [x] No

## How Has This Been Tested?

All new code has been covered by added unit tests; coverage is 96% for `asset_groups.py` and 95% for `devices.py`.  Runtime log of live test against DEV01 is here: [member_functions_asset_groups.log](https://github.com/carbonblack/carbon-black-cloud-sdk-python/files/13678033/member_functions_asset_groups.log)


src/cbc_sdk/platform/asset_groups.py
```diff
@@ -94,6 +97,104 @@ def _query_implementation(cls, cb, **kwargs):
         """
         return AssetGroupQuery(cls, cb)
 
+    def list_member_ids(self, **kwargs):
```
avanbrunt-cb, line 100: Why kwargs instead of parameters with default values? Also the name filter is confusing and requires you to read the description would recommend `membership`

```
def list_member_ids(self, start=0, rows=20, membership="ALL"):
```

src/cbc_sdk/platform/asset_groups.py
```diff
@@ -94,6 +97,104 @@ def _query_implementation(cls, cb, **kwargs):
...
+        if filt not in AssetGroup.VALID_MEMBER_FILTERS:
+            raise ApiError(f"invalid filter value: {filter}")
+        query_params = {}
+        for name in ("rows", "start"):
+            if name in kwargs:
+                query_params[name] = kwargs[name]
+        member_data = self._cb.get_object(self._build_api_request_uri() + "/members", query_params)
+        id_list = []
+        if filt == "ALL":
+            id_list = [int(m["external_member_id"]) for m in member_data["members"]]
```
avanbrunt-cb, line 131: What if I want to know what the membership type is where an asset group has both dynamic and manual memberships? Instead of a list of ids should we return the entire `members` list

src/cbc_sdk/platform/devices.py
```diff
@@ -315,6 +318,124 @@ def nsx_remediation(self, tag, set_tag=True):
             return None  # clearing tag is a no-op in this case
         return NSXRemediationJob.start_request(self._cb, self.id, tag, set_tag)
 
+    def get_asset_group_ids(self, **kwargs):
```
avanbrunt-cb, line 321: Why are we using kwargs for a function with a single keyword?

src/cbc_sdk/platform/devices.py
```diff
@@ -315,6 +318,124 @@ def nsx_remediation(self, tag, set_tag=True):
...
+                          Default is "ALL".
+
+        Returns:
+            list[AssetGroup]: A list of asset groups this device belongs to.
+        """
+        rc = Device.get_asset_groups_for_devices(self._cb, self, **kwargs)
+        return [self._cb.select("AssetGroup", v) for v in rc[str(self._model_unique_id)]]
+
+    @staticmethod
+    def _normalize_member_list(member_list):
```
avanbrunt-cb, line 366: Should this be a more generic helper function in the CBC SDK to normalize a list and you can specify the key to pull out of an object?

src/cbc_sdk/platform/devices.py
```diff
@@ -315,6 +318,124 @@ def nsx_remediation(self, tag, set_tag=True):
...
+                return_list.extend(Device._normalize_member_list(m))
+            else:
+                try:
+                    return_list.append(str(int(m)))
+                except ValueError:
+                    pass
+        return return_list
+
+    @classmethod
+    def get_asset_groups_for_devices(cls, cb, *args, **kwargs):
```
avanbrunt-cb, line 401: This is an advanced and complex function where as our target audience for the SDK is more beginner to moderate users. Why complex the interface? What problem are we trying to solve?

avanbrunt-cb (commented):
There appears to be a design pattern change here for helper functions. Can you provide context on why this is necessary and how it improves the development experience? 

This would break auto code complete give args and kwargs is not functionality defined only documentation defined

abowersox-cb:
Withdrawn until I can simplify matters.

PR #470, opened 2024-01-16
CBAPI-4252: Asset Groups: Get Effective Policy for Device

abowersox-cb:
## Pull request checklist

https://jira.carbonblack.local/browse/CBAPI-4252

## Pull Request Description

Added many functions to `AssetGroup`, `Device`, and `Policy` that implement the various preview operations.

- [ ] Yes
- [x] No

## How Has This Been Tested?

Unit tests have been created for all new functions; coverage is at 99% for `asset_groups.py`, 96% for `devices.py`, and 97% for `policies.py`.  Log file from live run is here: [preview_funcs_changes.log](https://github.com/carbonblack/carbon-black-cloud-sdk-python/files/13954974/preview_funcs_changes.log)

### DIFF
--- a/src/cbc_sdk/platform/devices.py
+++ b/src/cbc_sdk/platform/devices.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 # *******************************************************
-# Copyright (c) VMware, Inc. 2020-2023. All Rights Reserved.
+# Copyright (c) VMware, Inc. 2020-2024. All Rights Reserved.
 # SPDX-License-Identifier: MIT
 # *******************************************************
 # *
--- a/src/cbc_sdk/platform/devices.py
+++ b/src/cbc_sdk/platform/devices.py
@@ -492,7 +492,7 @@ class Device(PlatformModel):
         elif isinstance(devices, Device):
             device_ids = [str(devices.id)]
         else:
-            device_ids = Device._collect_devices(devices)
+            device_ids = [str(v) for v in Device._collect_devices(devices)]
         if len(device_ids) > 0:
             postdata = {"external_member_ids": device_ids}
             if membership != "ALL":
--- a/src/cbc_sdk/platform/devices.py
+++ b/src/cbc_sdk/platform/devices.py
@@ -445,6 +445,25 @@ class Device(PlatformModel):
         return Device.preview_remove_policy_override_for_devices(self._cb, [self])
 
     @classmethod
+    def _collect_devices(cls, devices):
+        """
+        Collects a list of devices as IDs.
+
+        Args:
+            devices (list): A list of items, each of which may be either integer device IDs or ``Device`` objects.
+
+        Returns:
+            list[int]: A list of integer device IDs.
+        """
+        device_ids = []
+        for d in devices:
+            if isinstance(d, Device):
+                device_ids.append(d.id)
+            elif isinstance(d, int):
+                device_ids.append(d)
+        return device_ids
+
+    @classmethod
     def get_asset_groups_for_devices(cls, cb, devices, membership="ALL"):
         """
         Given a list of devices, returns lists of asset groups that they are members of.
@@ -468,17 +487,12 @@ class Device(PlatformModel):
         """
         if membership not in Device.VALID_ASSETGROUP_FILTERS:
             raise ApiError(f"Invalid filter value: {membership}")
-        device_ids = []
         if isinstance(devices, int):
             device_ids = [str(devices)]
         elif isinstance(devices, Device):
             device_ids = [str(devices.id)]
         else:
-            for d in devices:
-                if isinstance(d, int):
-                    device_ids.append(str(d))
-                elif isinstance(d, Device):
-                    device_ids.append(str(d.id))
+            device_ids = Device._collect_devices(devices)
         if len(device_ids) > 0:
             postdata = {"external_member_ids": device_ids}
             if membership != "ALL":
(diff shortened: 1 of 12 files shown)


kebringer-cb (approved):

src/cbc_sdk/platform/devices.py
```diff
@@ -414,6 +415,54 @@ def remove_from_groups(self, groups):
         if len(actual_groups) > 0:
             self._refresh()
 
+    def preview_add_policy_override(self, policy_id):
```
avanbrunt-cb, line 418: I'm a bit confused by this as a policy_override for a device would always result in the policy being applied so what are we actually previewing here?

avanbrunt-cb: Remove policy override makes sense as we want to know what will be assigned but adding an override seems redundant

src/cbc_sdk/platform/policies.py
```diff
@@ -1031,10 +1032,28 @@ def preview_rank_change(self, new_rank):
                             where _N_ is the total number of policies in the organization.
 
         Returns:
-            list[PolicyRankChangePreview]: A list of objects containing data previewing the policy changes.
+            list[DevicePolicyChangePreview]: A list of objects containing data previewing the policy changes.
         """
         return Policy.preview_policy_rank_changes(self._cb, [(self._model_unique_id, new_rank)])
 
+    def preview_add_policy_override(self, devices):
```
avanbrunt-cb, line 1039: Not sure this is necessary only remove policy override applies

src/cbc_sdk/platform/devices.py
abowersox-cb: If you think it's not needed, I'll just ditch that particular helper.  It can be erased easily without affecting anything else.

src/cbc_sdk/platform/policies.py
abowersox-cb: This previews what happens to a series of devices if you add _this_ policy as an override for them.  The results could be different for each device, as to _which_ policy is being overridden.

However, if you think this is also redundant, this can be erased with little problem.

src/tests/unit/fixtures/platform/mock_policies.py
```diff
@@ -2543,3 +2543,29 @@
...
+    "preview": [
+        {
+            "current_policy": {
+                "id": 11200,
+                "position": 2
+            },
+            "new_policy": {
+                "id": 65536,
+                "position": 1
+            },
```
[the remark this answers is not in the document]
abowersox-cb: That's how this data works. The current policy (11200) is in position 2, and it's being overridden by the new policy (65536) in position 1.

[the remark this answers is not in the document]
abowersox-cb: And which devices are being changed in this fashion is specified by the `asset_query` element.

src/cbc_sdk/platform/devices.py
avanbrunt-cb: Yeah it doesn't seem useful as it will only show the change you are making and won't have any unknown discoverable impacts

src/cbc_sdk/platform/policies.py
avanbrunt-cb: I guess I'm confused when would a policy override not apply as it should out rank any policy based position

src/tests/unit/fixtures/platform/mock_policies.py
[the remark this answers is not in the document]
avanbrunt-cb: I'm confused though as the concept is not to make a policy position change but rather a policy override change for a device. Why is the policy position changing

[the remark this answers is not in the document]
avanbrunt-cb: An override is a manual policy assignment not a policy position change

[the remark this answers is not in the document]
abowersox-cb: OK, suppose your current effective policy is policy Alpha with a priority of 2. You assign a new policy as an override, Bravo with priority 4.  You get a preview back that has Alpha and 2 under "current," Bravo and 4 under "new."

If you then remove the override later, you would get a preview with Bravo and 4 under "current," and Alpha and 2 under "new."

I took these data elements straight from the return data I got from Postman.

src/cbc_sdk/platform/policies.py
abowersox-cb: Well, let's say one device has policy Alpha with priority 2, and another device has policy Bravo with priority 3. If you then have policy Charlie with priority 4 override both devices, you would get two separate preview notifications, both having Charlie as the "new" policy, but one with Alpha as "current," and one with Bravo as "current."

src/cbc_sdk/platform/devices.py
abowersox-cb: Removing it, and patching the appropriate unit test to point to `Device.preview_add_policy_override_for_devices`.

src/tests/unit/fixtures/platform/mock_policies.py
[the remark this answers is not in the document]
abowersox-cb: The return data format is the exact same for all the "preview" functions, whether it's priority changes, policy overrides, group membership changes, group adds, or group deletes.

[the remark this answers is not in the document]
avanbrunt-cb: Yeah that makes sense just seems like a weird API design now that I think about it for policy overrides

avanbrunt-cb (approved):
Overall really great and seems useful. Let's see if people use it and have feedback. Still curious as to how many people will make changes from a script now that asset groups is out.

PR #471, opened 2024-01-17
added additional unit test to device to validate inheritance of to_js…

heiderje-vmware:
## Pull request checklist

[CBAPI-5059](https://jira.carbonblack.local/browse/CBAPI-5059)

## Pull Request Description
Adds inheritable to_json() to NewBaseModel and removes replaces _info in rtd

- [ ] Yes
- [x] No

## How Has This Been Tested?
unit tests and live test against devices via the sdk

### DIFF
--- a/src/cbc_sdk/base.py
+++ b/src/cbc_sdk/base.py
@@ -773,6 +773,15 @@ class NewBaseModel(object, metaclass=CbMetaModel):
 
         return self._cb.select(join_cls, field_value)
 
+    def to_json(self):
+        """
+        Return a json object of the response.
+
+        Returns:
+            Any: The returned attribute value.
+        """
+        return self._info
+
 
 class UnrefreshableModel(NewBaseModel):
     """Represents a model that can't be refreshed, i.e. for which ``reset()`` is not a valid operation."""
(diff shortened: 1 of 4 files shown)


avanbrunt-cb (approved):

src/cbc_sdk/base.py
```diff
@@ -773,6 +773,15 @@ def _join(self, join_cls, field_name):
...
         return self._cb.select(join_cls, field_value)
 
+    def to_json(self):
+        """
+        Return a json object of the response.
+
+        Returns:
+            Any: The response dictionary representation.
+        """
+        return self._info
```
abowersox-cb, line 783: I'd kind of rather have it return `copy.deepcopy(self._info)` instead, so you het a clean copy that doesn't affect the internal state of the object if you try and change it.

avanbrunt-cb: That's fair though make sure to do the same thing for any other to_json overrides

abowersox-cb: I absolutely agree with that!

abowersox-cb (approved):

PR #500, opened 2024-04-17
CBAPI-5258: Implement audit log search API

abowersox-cb:
## Pull request checklist

https://jira.carbonblack.local/browse/CBAPI-5258

## Pull Request Description

Implemented audit log search by using a query object in the standard pattern. Query object supports standard criteria and exclusions and async queries.

- [ ] Yes
- [x] No

## How Has This Been Tested?

Unit tests have been written; test coverage at 97%.


src/cbc_sdk/platform/audit.py
```diff
@@ -39,3 +67,275 @@ def get_auditlogs(cb):
...
+        self._sortcriteria = {}
+        self._search_after = None
+        self.num_remaining = None
+        self.num_found = None
+        self.max_rows = -1
+
+    @staticmethod
+    def _create_valid_time_filter(kwargs):
+        """
+        Verifies that an alert criteria key has the timerange functionality
```
emitreva-cb, line 117: `alert criteria` ?

abowersox-cb: Yeah, that's why this was still only a draft PR, because I wanted one more chance to go over all the docstrings. :)

src/cbc_sdk/platform/__init__.py
```diff
@@ -8,7 +8,9 @@
 
 from cbc_sdk.platform.alerts import Alert as BaseAlert
 
-from cbc_sdk.platform.audit import AuditLog
+from cbc_sdk.platform.audit import AuditLogRecord
+
+from cbc_sdk.platform.audit import AuditLogRecord as AuditLog
```
avanbrunt-cb, line 13: Why do we need to rename this?

src/cbc_sdk/platform/audit.py
```diff
@@ -39,3 +67,275 @@ def get_auditlogs(cb):
         """
         res = cb.get_object("/integrationServices/v3/auditlogs")
```
avanbrunt-cb, line 68: Is there a plan to replace this with `_queue`?

src/cbc_sdk/platform/__init__.py
abowersox-cb: Because the name of the class is misleading; it was `AuditLog` but it really represents a single _record_ in the audit log, hence, `AuditLogRecord`. The old name must be kept around, however, for backward compatibility.

src/cbc_sdk/platform/audit.py
abowersox-cb: Yes, that would be story CBAPI-5259.

src/cbc_sdk/platform/__init__.py
avanbrunt-cb: AuditLog is both plural and singular. It is an Audit Log

abowersox-cb: So, you're saying that the class _shouldn't_ be renamed?

avanbrunt-cb: Yes I am suggesting we don't rename the class unless we have a functional reason why we need a different name

avanbrunt-cb (approved):

PR #504, opened 2024-04-23
CBAPI-5260: Implement Audit Log Export

abowersox-cb:
## Pull request checklist

https://jira.carbonblack.local/browse/CBAPI-5260

## Pull Request Description

Implemented the `export()` function on `AuditLogQuery`, analagous to other exports that are used elsewhere.

- [ ] Yes
- [x] No

## How Has This Been Tested?

Unit tests have been added; code coverage stands at 97% on `audit.py`.

### DIFF
--- a/src/cbc_sdk/platform/audit.py
+++ b/src/cbc_sdk/platform/audit.py
@@ -393,6 +393,11 @@ class AuditLogQuery(BaseQuery, QueryBuilderSupportMixin, CriteriaBuilderSupportM
         The actual results are retrieved by waiting for the resulting job to complete, then calling one of the methods
         on ``Job`` to retrieve the results.
 
+        Example:
+            >>> audit_log_query = cb.select(AuditLog).add_time_criteria(range="-1d")
+            >>> audit_log_export_job = audit_log_query.export(format="csv")
+            >>> results = audit_log_export_job.await_completion().result()
+
         Args:
             format (str): Format in which to return results, either "csv" or "json".  Default is "csv".
 
--- a/src/cbc_sdk/platform/audit.py
+++ b/src/cbc_sdk/platform/audit.py
@@ -382,3 +382,6 @@ class AuditLogQuery(BaseQuery, QueryBuilderSupportMixin, CriteriaBuilderSupportM
         result = resp.json()
         results = result.get("results", [])
         return [self._doc_class(self._cb, item) for item in results]
+
+    def export(self):
+        pass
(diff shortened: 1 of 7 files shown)


avanbrunt-cb (approved):

src/cbc_sdk/platform/audit.py
```diff
@@ -382,3 +385,27 @@ def _run_async_query(self, context):
...
+        """
+        if format not in AuditLogQuery.VALID_EXPORT_FORMATS:
+            raise ApiError(f"invalid export format '{format}'")
+        url = self._build_url("/_export")
+        request = self._build_request(0, -1)
+        request["format"] = format
+        resp = self._cb.post_object(url, body=request)
+        result = resp.json()
+        if "job_id" in result:
+            return Job(self._cb, result["job_id"])
```
kebringer-cb, line 410: The work around done on the Jobs service for Devices when the jobs /progress API returns this error instead of Complete needs to be applied generally to all uses of jobs progress.  The same error occurs on Audit Log.

{"error_code":"PROGRESS_NOT_FOUND","message":"No progress available for job 13656055","job_id":13656055}

src/cbc_sdk/platform/audit.py
```diff
@@ -382,3 +385,27 @@ def _run_async_query(self, context):
         result = resp.json()
         results = result.get("results", [])
         return [self._doc_class(self._cb, item) for item in results]
+
+    def export(self, format="csv"):
```
kebringer-cb, line 389: An example please so it's easy to use.
audit_log_query = cb.select(AuditLog).add_time_criteria(range="-1d")
audit_log_export_job = audit_log_query.export(format="csv")
results = audit_log_export_job.await_completion().result()

src/cbc_sdk/platform/audit.py
abowersox-cb: Are you saying that _all_ uses of the `/progress` API for jobs should be considered "broken" from here on out? Or are audit log jobs just broken in the same way that device jobs are?  There's an exception mechanism in the `Job` class I can tweak to make it use the "status" API for waiting for certain types of jobs; I can employ that here.

kebringer-cb: yes, jobs progress is broken for all uses.  Once a job completes, the /progress endpoint returns and error.

abowersox-cb: OK, then I guess the answer is to take out the `wait_status` determination code in `Job` and just wire it up as if `wait_status` were always `True`.  So the old "progress" code will disappear entirely.

abowersox-cb: The above has been done.

src/cbc_sdk/platform/audit.py
abowersox-cb: Thanks for providing the example text. Added.

kebringer-cb (approved):

